PR #36, opened 2016-10-03
wiki: link added

aj07:
Link for http://deeplearning.net/software/theano/ should be added.
Also, please add label for hackoberfest.


chewxy:
Merged

PR #43, opened 2016-10-09
Simplifies func returns & creates better Solver struct

ArxdSilva:


chewxy:
hi,

is there a reason everything got stuffed into a Solver struct ? I'm curious about your thought processes

Issue #50, opened 2016-10-18
Readme example differentiation typo


HeinOldewage:
The code only prints the value if an error occurred. The ifs should test for err == nil.

    if xgrad, err := x.Grad(); err != nil {
        fmt.Printf("dz/dx: %v", xgrad)
    }

    if ygrad, err := y.Grad(); err != nil {
        fmt.Printf("dz/dy: %v", ygrad)
    }

chewxy:
Right you are! Could you make a PR? I'm abroad atm, but can merge pull requests

PR #51, opened 2016-10-20
Updated README example for readability

3ygun:
Wanted:
- Consistent `if err != nil {` styles. (Lines 76-77 now match 82-83)
- Copy, Paste, Install & Run: so added the missing `package main`.

As I was confused when reading through.

**Update**: Noticed there were other example on the README let me know if I should update those to this format also.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,8 @@ Whilst Gorgonia's implementation doesn't enforce the separation of thought as fa
 Here's an example - say you want to define a math expression `z = x + y`. Here's how you'd do it:
 
 ```go
+package main
+
 import (
 	"fmt"
 	"log"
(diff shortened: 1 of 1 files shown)


chewxy:
Fixes #50

chewxy:
Some changes were made to the printing of the `*dualValue` type, which results in the result of the Theano-like example that you see.

Specifically:

```
dz/dx: 1 | 1 :: Float64{1}
dz/dy: 1 | 1 :: Float64{1}
```

The value is `1` and the value of the gradient is `1` of type `const Float64`. I'm still playing around with the reporting of values, so my original example should suffice.

chewxy:
Thank you btw

3ygun:
@chewxy No problem! Look forward to using **gorgonia**. Sorry about not commenting about the printing had to go to dinner 🌮

PR #52, opened 2016-10-27
Stack method

### LINKED ISSUE #16: Add Stack() to Tensor
It'd be like Numpy's stack.

Preliminary design looks something like this:

`func (t *Tensor) Stack(other *Tensor, axis int) (*Tensor, error) `

And a package-level function:

`func Stack(axis int, ...*Tensor) (*Tensor, error)`

chewxy:
Fixes #16


chewxy:
Failure at go tip is probably due to this: https://github.com/golang/go/issues/17596

Will wait a couple of days before merging

Issue #46, opened 2016-10-11
Rework errors


chewxy:
This task is broken into two parts:
- Wrap all errors in Gorgonia with the [errors](https://github.com/pkg/errors) package, with meaningful error messages.
- Remove all of the `runtime.Caller()` function from the basic errors.

robjloranger:
i'd like to take a shot at this

chewxy:
Added you to the list of collaborators. It's just so I can assign the issue. Please use a personal fork and send a pull request

robjloranger:
Awesome, will do. Thanks 😃

robjloranger:
I haven't had time to wrap my head around this. I'll keep looking back to it but if anyone else wants to take it on please do.

chewxy:
Cool. thank you Rob

NDari:
I have started work on this Issue. The WIP is in the PR above.

chewxy:
Thank you. I'll add you to the list of assignees

chewxy:
This should have been closed.

PR #59, opened 2016-11-04
[WIP] work on getting the gorgonia to use the errors package

NDari:
This is the work in progress for the integration of the errors package into gorgonia which addressed #46 

As this work touches a large surface area, I wanted to open this branch early to mark the progress, so that I can get some feedback (if any) about the approach.

once all the work is done, I will be squashing all the commits into one with a meaningful commit message to keep everything in master clean, so please note that I may be doing a force push at some point.


chewxy:
Hi Naseer, excellent work here so far! 

I was wondering if you think putting the error in the `Error` struct provided by Gorgonia might be a better idea? 

Example in comments in differentiation.go

NDari:
@chewxy You are correct! That is a decision I made for now, since I was not sure what value the error type added, beyond what the `errors` package provided. For example I didnt see where some decision was being made based on the type of the error to do something for debugging or something else. Based on that (and please correct me if I am wrong), I went ahead with the idea that I could remove the extra error types and fully rely on the error package alone. 

If we want to keep the type, I may be able to rework the `ErrorType` structure to make it play nice with the errors package somehow though I dont know exactly how right now. Let me know what you think.

chewxy:
Nah it's not necessary. The only reason I first added error types was to help trace the errors in the early stages of writing this (I had wanted to know the "class" of errors that I needed to work on).

I think you are on the right track

NDari:
Sounds good.

NDari:
@chewxy This is a bit strange to me, so I wonder if you can take a quick look and see if everything looks ok to you. in [nn.go line 53](https://github.com/chewxy/gorgonia/blob/master/nn.go#L53) we have:

```go
	if retVal, err = Log(omo); err != nil {
		err = errors.Wrap(err, operationError)
	}

	if retVal, err = HadamardProd(omt, retVal); err != nil {
		err = errors.Wrap(err, operationError)
	}

	if retVal, err = Add(tLogO, retVal); err != nil {
		err = errors.Wrap(err, operationError)
	}

	return Neg(retVal)
```

should each of these if branches be returning? Or is the intent to simply check for the error, wrap it, and continue with the rest of the function (the rest of the if functions).

It really does not effect my work, but the change in the pattern caught my eyes, and unfortunately I am not that familiar with the code yet to figure out what the intent is easily.

chewxy:
Yeah. They should have returned early. You caught a bug too! 👍

NDari:
Since I am in here I will just fix that!

NDari:
@chewxy how do you feel about removing the named returned values from function definitions? for example:

```go
func add(x, y int) (result int, err error) { .... }
```

would become

```go
func add(x, y int) (int, error) { ... }
```

This is the recommended way to do it (see [here](https://tour.golang.org/basics/7) for example) and thats how the Go stdlib is written. More pragmatically, I find the logic to be a bit easier to follow when the return statements explicitly state what they are returning, and I do not need to go and hunt the state of the declared variabled at the top to see what is actually returned. I have found a few subtle bugs so far do to this which I can dig up.

I am getting a bit bogged down in my branch due to errors that are much harder to track due to these types of thing. If you dont have a strong preference, it would help me a great deal to make this change, and I would volunteer to do it myself :)

chewxy:
I think naked returns should only be applied to short functions. A lot of the functions in `operations.go`, I guess, qualify for "short function" given that they're just a redirect mostly to `applyOp`

NDari:
I just took a look at those functions and I agree for most of those functions, a naked return is certainly doable. `applyOp` itself is a bit longish though, I think.

We can certainly table this issue for now. I can remove the naked return in the more complex function for my personal branch and put them back in after I am familiar with the codebase, but will not need to check that in.

op_math.go
```diff
@@ -362,17 +355,15 @@ func (op elemBinOp) IncrDo(incr Value, inputs ...Value) (err error) {
...
 
 		add := newEBOByType(addOpType, incr.Type(), retVal.Type())
 		if retVal, err = add.UnsafeDo(incr, retVal); err != nil {
-			err = errors.Wrapf(err, unsafeDoFail, add)
-			return
+			return errors.Wrapf(err, unsafeDoFail, add)
 		}
-		err = noIncrErr{retVal}
-		return
+		// return errors.New(incrErr) //////////////////// is this ok????
```
chewxy: Nope. Not OK.  IIRC, there are parts which rely on an `error` that has a `Value()` method, for graceful error handling.

op_math.go
```diff
@@ -424,13 +415,11 @@ func (op elemUnaryOp) Type() Type {
 
 func (op elemUnaryOp) inferShape(retType Type, inputs ...*Node) (retVal types.Shape, err error) {
 	if len(inputs) != 1 {
-		err = NewError(GraphError, "Pointwise unary operations only take ONE input. Got %d inputs instead", len(inputs))
-		return
+		return nil, errors.Errorf(unaryOpFail, len(inputs))
```
chewxy: I think you can skip any 

`if len(inputs) != 1` type of error fixing, because there is a branch of code that removes this check entirely

operatorPointwise_binary.go
```diff
@@ -422,26 +409,14 @@ func hadamardProdDiff(x, y, z *Node) (err error) {
...
 	err = mul.IncrDo(xdv.d, ydv.Value, zdv.d)
 	if err != nil {
-		var ver Valuer
-		var ok bool
-		if ver, ok = err.(Valuer); !ok {
-			return
-		}
-
-		xdv.SetDeriv(ver.Value()) // ignore errors on purpose
+		return errors.Wrap(err, "IncrDo call failed")
```
chewxy: This will cause A LOT of bugs. Don't do this. 

The main idea is this: there are two `Value` types: `Tensor` and `Scalar`.

`IncrDo` is the representation of ` x += (op)...`. It relies on the fact that Tensors are generally pointer types.

Now, you can't do the same on scalar types because they ain't pointers, so you have to wait for an error, see if it is a `Valuer`, and if it isn't then use `SetDeriv` to update the increased version

op_math.go
NDari: I was worried about this. Will revert.

chewxy:
Oh wow you are a fast commenter. yeah, that will be great. Thank you.

operatorPointwise_binary.go
NDari: I have noticed that my scalar*tensor and tensor*scalar ops are now failing... I will revert this.

op_math.go
NDari: Is that code coming up in a PR? are you recommending that I revert this?

NDari:
Ha! I am working on this right now... i am usually not this fast ;)

chewxy:
FYI you can skip updating any code that looks like this:

`go
if len(inputs) != ...{

}`

The reason is this: https://github.com/chewxy/gorgonia/blob/3b7b07e44e5f39d665a8197c8a68569b30f990ab/utils.go#L280  (it will be updated to match your errors :))

op_math.go
chewxy: yes please. It will come up in a PR after yours :)

NDari:
Requested changes have been made. all the arity checks have been reverted back to their original state. I am not sure how to "accept" your review besides dismissing it.

chewxy:
Yeah, it just shows up as "outdated" in my github. All good. Tell me when you think this is ready for merging.

differentiation.go
```diff
@@ -164,22 +163,19 @@ func Backpropagate(outputs, gradOutputs, wrt Nodes) (retVal Nodes, err error) {
 
 	var sortedNodes Nodes
 	if sortedNodes, err = Sort(g); err != nil {
-		err = errors.Wrap(err, sortFail)
-		return
+		return nil, errors.Wrap(err, sortFail)
```
chewxy: I think this may be better:

```go
return nil, errors.Wrapf(err, sortFail, "Backprop")
```

This way, the user doesn't have to do the whole `Cause` thing to quickly figure out where the error happened. Thoughts?

dual.go
```diff
@@ -229,9 +230,10 @@ func dvBind(op Op, inputs []*dualValue) (retVal *dualValue, err error) {
 
 	var ret Value
 	if ret, err = op.Do(vals...); err == nil {
-		retVal = dvUnit(ret)
+		return dvUnit(ret), nil
+	} else {
+		return nil, errors.Wrap(err, opDoFail)
```
chewxy: perhaps wrap the `Op` name into the errors as well? (all `Op`s implement  `fmt.Stringer`)

chewxy:
FYI, I added Coveralls. You may want to fetch and merge from Master.

dual.go
NDari: I have the same feelings as what I have put on the previous issue. let me know what you think about that!

differentiation.go
chewxy: OK. Your idea is superior

NDari:
I have rebased and will push soon. tonight is election night so I dont know if ill get much done. but I plan to get the lowest level done by this (american) weekend. I will rebase and squash all the commits into one and let you know when that is ready!

dual.go
chewxy: actually, I think the op name won't show up. This is because in some complex operations (say `softmax`, there would be multiple ops in the `softmax` function. If `%+v` is used, it will only show that the error happens in `SoftMax()`

chewxy:
Yeah you should prolly go vote.

dual.go
NDari: You are right, I was conflating the function/method with the mathematical operation(s) inside a given function. I still think with the given line numbers for a stack trace, it may be good enough to debug whatever issues may be needed. If you think this is still something that adds value, can you open a new issue? I would like to get this out of the door as is without too much more scope creep, unless you think otherwise

chewxy: OK. I'll add a new issue. btw, see your linked in. I asked for your email :)

NDari:
Oh I voted last weekend. I am going to watch TV hoping that the world does not burn tonight.

dual.go
NDari: didnt you get my reply? I almost replied right away!

NDari:
@chewxy can you look at this: in operations, line 55 we have:

```go
	if s, err = op.inferShape(retType, children...); err == nil {
		typeSysLogf("inferred type: %v", retType)
		shapeLogf("inferred shape %v", s)
		retVal = newUniqueNode(withType(retType), withOp(op), withChildren(children), withGraph(g), WithShape(s...))
	} else {
		retVal = newUniqueNode(withType(retType), withOp(op), withChildren(children), withGraph(g))
	}
	return
```

The else branch happens when `err != nil`, right? in that case, we wouldnt we just want to `return nil, errors.wrap(err, ....)` here?

chewxy:
That `else` branch is a vestige from a time when the shape inference was relegated to runtime (so at runtime, the input matrices will determine the shape of the nodes)...

As such you may note that it will create a node without a shape parameter. The error from the `InferShape` methods just tells us if the shape inference succeeded. The node should still be created either way.

That said, by and large, the runtime shape inference feature has been removed, mainly due to the pain I faced when writing slightly more complex neural networks. However, I think that runtime shape inference could be a useful feature (Theano has only runtime shape inference). 

Not sure what the best foot forwards is. This is the dilemma:

- Compile time shape inference pros: no runtime errors with shapes - all the shapes are known before hand
- Compile time shape inference cons: makes it harder to write dynamic computation graphs.
- Run time shape inference pros: easy to write dynamic computation graphs.
- Run time shape inference cons: Absolutely shitty subtle errors will happen that will make you pull your hair out, especially with matrices that have compatible shapes when performing ops.

I'm thinking for the mean time, leave it as it is? I'm still thinking about runtime shape inference. Some input from you would be nice.

NDari:
for the problem at hand, I will leave it as is for now. We can always extend/enhance the errors down the line. This is just an easy first pass (and it helps me to know what the hell is going on :P)

As for your questions, I am not in any position to talk about this at all. I dont have a good understanding of what class of problems are easier/better solved with dynamic graphs. However, just approaching Gorgonia as a business (which it is not), I think time is best spent on features and enhancements over debugging things for both you and the consumers of this library. So, again, not knowing the use cases between dynamic and static computational graphs, I would err on the side of not messing with something that could potentially bog you down with hair loss.

NDari:
@chewxy The work is done. I will be opening a few issues this evening US time for some future cleanup work. One big item remaining is actually handing and displaying these errors: I have some ideas which i will post in one of the issues. There is also some spit and polish that we can do to make this whole thing a better experience, and there is a lot of redundancies in the error messages. etc etc. sooo much work to be done, so ill leave it up to you to decide the priorities of these things.

Finally, I have squashed the whole thing into one commit to keep the master branch clean. Let me know how you like this whole thing.

chewxy:
Merged. Thank you. Good job. I like it.

PR #93, opened 2017-02-25
Cuda2

chewxy:
CUDA branch v2: now with batched CUDA calls... up to 3 times faster!

PR #99, opened 2017-03-22
Masked tensor

kabaka0:
As promised, set about trying to implement basic masked array functionality.

To begin with, created a new iterator type 'MultIterator', which is designed to iterate over multiple arrays simultaneously, with the same syntax as 'FlatIterator', to allow switching between them (it uses an array of FlatIterators internally). For single non-masked arrays, MultIterator is about 20% slower than FlatIterator. However, it only calculates offsets for unique shapes/stride combinations, and so when indexing arrays of same shape, a single FlatIterator is shared between them all, allowing significant compute savings. 

```
func BenchmarkFlatIteratorMulti6(b *testing.B) {
	ap := make([]*AP, 6)
	for j := 0; j < 6; j++ {
		ap[j] = NewAP(Shape{30, 60, 10}, []int{1000000, 15000, 50})
	}
	it := NewMultIterator(ap...)
	for n := 0; n < b.N; n++ {
		for _, err := it.Next(); err == nil; _, err = it.Next() {
		}
		it.Reset()
	}
	DestroyMultIterator(it)
}
```

You could  create a MultiIterator from tensors directly

```
T1 := New(Of(Float64), WithShape(3, 20), WithMaskStrides([]bool{true, true}))
T2 := New(Of(Float64), WithShape(3, 20), WithMaskStrides([]int{20,1}))
T3 := New(Of(Float64), FromScalar(7))
it := MultIteratorFromDense(T1, T2, T3)
```

It also means that you don't have to worry when creating functions of multiple arguments in which the same array could be repeated as different arguments - in which case naive use of FlatIterator could cause that array to be iterated multiple times in a single for loop iteration - with MultIterator that can not happen. 

As for the mask, for the time being I opted to simply add a []bool to the Dense struct, and an additional stride int to AP. MultIterator supports masked operations, such as NextValid() or NextInvalid(), in addition to Next(). Examples of usage can be seen in dense_maskmethods_test.go and iterator_test.go.

```
func TestMaskedIteration(t *testing.T) {
	assert := assert.New(t)
	T := New(Of(Float64), WithShape(2, 3, 4, 5))
	assert.True(len(T.mask) < 1)
	dataF64 := T.Data().([]float64)
	for i := range dataF64 {
		dataF64[i] = float64(i)
	}
	for i := 0; i < 5; i++ {
		T.MaskedEqual(float64(i) * 10.0)
	}

	it := MultIteratorFromDense(T)

	j := 0
	for _, err := it.Next(); err == nil; _, err = it.Next() {
		j++
	}
	it.Reset()
	assert.True(j == 120)

	j = 0
	for _, err := it.NextValid(); err == nil; _, err = it.NextValid() {
		j++
	}
	it.Reset()
	assert.True(j == 115)

	j = 0
	for _, err := it.NextInvalid(); err == nil; _, err = it.NextInvalid() {
		j++
	}
	it.Reset()
	assert.True(j == 5)
}
```

I did not want to spend too much time going further before agreeing on the basics. While I show some basic mask setting  operations in dense_maskmethods.go, I only do this for float64 tensors as a demonstration - the functionality would have to be implemented in genlib at some point which would take me sometime to do properly as this is my first time using text/template.

I also did not optimize masked iteration, there are smarter ways to find next valid/valid, e.g. by processing >=8 bytes at once, but I figure that it's best to leave that until the structure is agreed upon.


chewxy:
Code looks good. Thank you for adding tests and bench.

I like the AP an Dense addition - it sounds good. Do you need help with genlib?

chewxy:
btw, I won't merge this yet until you feeel it's ready to be merged.

Also, please add your full name under SIGNIFICANT CONTRIBUTORS

chewxy:
ADDITIONAL IDEAS HERE (to be added when I'm not awake at 4.20 am)

tensor/iterator_mult.go
```diff
@@ -0,0 +1,295 @@
+package tensor
+
+// MultIterator is an iterator that iterates over multiple tensors, including masked tensors.
+//  It utilizes the *AP of a Tensor to determine what the next index is.
+// This data structure is similar to Numpy's flatiter, with some standard Go based restrictions of course
+// (such as, not allowing negative indices)
+
```
chewxy: delete empty line - fixes godoc issues

tensor/iterator_mult.go
```diff
@@ -0,0 +1,295 @@
...
+		for i, tt := range tts {
+			masks[i] = tt.mask
+		}
+		it.masks = masks
+	}
+	ReturnAPList(aps)
+	return it
+}
+
+// DestroyMultIterator creates a new MultIterator from a list of dense tensors
```
chewxy: maybe unexport this,

Then in any function that creates `MultIterator` add `runtime.SetFinalizer` to call this function?

tensor/shape.go
```diff
@@ -22,7 +22,9 @@ func (s Shape) TotalSize() int {
 	return ProdInts([]int(s))
 }
 
-func (s Shape) calcStrides() []int {
+// calcStrides has an additional optional argument, maks. It is used to mask out given dimensions
```
chewxy: I think there should be another function called `calcStridesWithMask`. `calcStrides` is called A LOT, might make sense to just split the functions

tensor/ap_test.go
```diff
@@ -385,7 +445,9 @@ func TestUntransposeIndex(t *testing.T) {
 	}
 }
 
-func TestBroadcastStrides(t *testing.T) {
+// TODO: This test fails, why?
```
chewxy: AFAICT, this shouldn't fail...

tensor/dense_io_test.go
```diff
@@ -1,3 +1,5 @@
+// +build ignore because not working
```
chewxy: io makes heavy use of `FlatIterator`. You should probably add some dense_io tests for the masked ones - what should masked tensors look like when printed?

tensor/dense_maskmethods.go
```diff
@@ -0,0 +1,127 @@
...
+	return maskedCmp(t, fn, val, 0.0)
+}
+
+//MaskedNotEqual masks an array where not equal to a given value
+func (t *Dense) MaskedNotEqual(val interface{}) error {
+	fn := func(x, y, z float64) bool { return x != y }
+	return maskedCmp(t, fn, val, 0.0)
+}
+
+//MaskedValues masks an array using floating point equality
```
chewxy: This entire section: :+1: I like the simplicity of it

tensor/iterator_mult.go
kabaka0: Noted. Will do.

tensor/ap_test.go
kabaka0: Not failing anymore. Must have written the comment in the middle of some changes and broken something.

tensor/shape.go
kabaka0: Sounds fine. I tried to stay away from long names but I agree with you that it is best to minimize optional arguments and empty interfaces, makes things less likely to break.

chewxy:
I'm more than happy with the general direction of things. It looks bloody good and adds A LOT of functionality to it.

As for genlib there is a general... structure to the templates:

```
switch t.t.Kind(){
	{{range .Kinds -}}
	{{if isParameterized . -}}
	{{else -}}
	case reflect.{{reflectKind .}}:
}
```

generates 

```
switch t.t.Kind(){
	case reflect.Int:
	case reflect.Uint:
	case reflect.Float32:
	case reflect.Float64:
	// etc as long as the type is a non-parameterized type
}
```

I've declared the properties of the types:

- isParameterized: Is the type parameterized? ([]T, *T, etc). Generally the `tensor` package doesn't handle these types. Which is why in the example above, there is a check and skip for parameterized types.
- isRangeable: Can there be a range of values? Applies to all number types, except `uintptr`,
- isSpecialized:   special consideration would be given to this type within a tensor - all numbers and string,
- isNumber: anything that can perform `+` `-` `*` `/`
- isFloat: need I say more?
- isEq: anything that is comparable - numbers and string and bool
- isOrd: anythihng that is comparable - numbers and string

There are a bunch of useful functions in genlib.go. The [`funcMap`](https://github.com/chewxy/gorgonia/blob/master/tensor/genlib/genlib.go#L114) are a list of functions available from inside the templates.

tensor/iterator_mult.go
kabaka0: Noted

kabaka0:
@chewxy 
So written a demo script in genlib, but I am having trouble running it. Currently I run  go install from within the genlib folder, and then running genlib, but keep getting error (with or without my new script)
`Go imports failed with exec: "goimports": executable file not found in $PATH for "../dense_getset.go"`
Clearly I am not doing this correctly. What is the correct build and test procedure?

chewxy:
Oh genlib wasn't created to run like most normal commands. I have always left that as a TODO. To run genlib you need to go to that directory, and `go run *.go`.  It will overwrite all the files automatically

kabaka0:
Thanks, I'll do that. Part of the problem was that I did not have goimports installed 🤦.

chewxy:
This works:

```go
package main

import (
	"fmt"
	"io"
	"reflect"
	"text/template"
)

var maskcmpMethods = []struct {
	Name    string
	Desc    string
	NumArgs int
	CmpFn   string
	Kinds   []reflect.Kind
}{
	{"MaskedEqual", "equal to ", 1, "x == y", nil},
	{"MaskedNotEqual", "not equal to ", 1, "x != y", nil},
}

const maskCmpMethodRaw = `// {{.Name}}(val) sets the mask to true where the corresponding data is {{.Desc}} val.append.append
// val must be the same type as the tensor
func (t *Dense) {{.Name}}(val1 interface{} {{if ge .NumArgs 2}}, val2 interface{}{{end}}  {{if ge .NumArgs 3}}, val3 ...interface{}{{end}} )(err error){
	if !t.IsMasked() {
		t.SetMaskStrides(t.strides)
		t.fix()
	}	
    it := MultIteratorFromDense(t)

    {{$numargs := .NumArgs}}
    {{$fn := .CmpFn}}
	switch t.t.Kind(){
	{{range .Kinds -}}
	{{if isParameterized . -}}
	{{else -}}
	case reflect.{{reflectKind .}}:
        data := t.Data().{{sliceOf .}}
        mask := t.mask
        x := val1.({{asType .}})
        {{if ge $numargs 2 -}} y := val2.({{asType .}})){{end}}
        {{if ge $numargs 3 -}} z := val3.({{asType .}})){{end}}
        for i, err := it.Next(); err == nil; i, err = it.Next() {
			j := it.LastMaskIndex(0)
			mask[j] = mask[j] || ({{$fn}})
		}
		it.Reset()
		destroyMultIterator(it)
    {{end}}
    {{end}}
}
return nil
}
`

var (
	maskCmpMethod *template.Template
)

func init() {
	maskCmpMethod = template.Must(template.New("maskcmpmethod").Funcs(funcs).Parse(maskCmpMethodRaw))
}

func maskcmpmethods(f io.Writer, generic *ManyKinds) {
	for _, mm := range maskcmpMethods {
		mm.Kinds = generic.Kinds
		maskCmpMethod.Execute(f, mm)
		fmt.Fprintf(f, "/* %s */ \n\n", mm.Name)

	}
}
```

This is the diff: 

``` diff
--- <unnamed>
+++ <unnamed>
@@ -3,6 +3,7 @@
 import (
 	"fmt"
 	"io"
+	"reflect"
 	"text/template"
 )
 
@@ -10,10 +11,11 @@
 	Name    string
 	Desc    string
 	NumArgs int
-	cmpFn   string
+	CmpFn   string
+	Kinds   []reflect.Kind
 }{
-	{"MaskedEqual", "equal to ", 1, "x == y"},
-	{"MaskedNotEqual", "not equal to ", 1, "x != y"},
+	{"MaskedEqual", "equal to ", 1, "x == y", nil},
+	{"MaskedNotEqual", "not equal to ", 1, "x != y", nil},
 }
 
 const maskCmpMethodRaw = `// {{.Name}}(val) sets the mask to true where the corresponding data is {{.Desc}} val.append.append
@@ -25,6 +27,8 @@
 	}	
     it := MultIteratorFromDense(t)
 
+    {{$numargs := .NumArgs}}
+    {{$fn := .CmpFn}}
 	switch t.t.Kind(){
 	{{range .Kinds -}}
 	{{if isParameterized . -}}
@@ -32,12 +36,12 @@
 	case reflect.{{reflectKind .}}:
         data := t.Data().{{sliceOf .}}
         mask := t.mask
-        x := val1.({{asType .}}))
-        {{if ge .NumArgs 2 -}} y := val2.({{asType .}})){{end}}
-        {{if ge .NumArgs 3 -}} z := val3.({{asType .}})){{end}}
+        x := val1.({{asType .}})
+        {{if ge $numargs 2 -}} y := val2.({{asType .}})){{end}}
+        {{if ge $numargs 3 -}} z := val3.({{asType .}})){{end}}
         for i, err := it.Next(); err == nil; i, err = it.Next() {
 			j := it.LastMaskIndex(0)
-			mask[j] = mask[j] || ({{.cmpFn}})
+			mask[j] = mask[j] || ({{$fn}})
 		}
 		it.Reset()
 		destroyMultIterator(it)
@@ -58,8 +62,9 @@
 
 func maskcmpmethods(f io.Writer, generic *ManyKinds) {
 	for _, mm := range maskcmpMethods {
-		argMethod.Execute(f, mm)
+		mm.Kinds = generic.Kinds
+		maskCmpMethod.Execute(f, mm)
 		fmt.Fprintf(f, "/* %s */ \n\n", mm.Name)
 
 	}
-}+}

```

Going thru them one by one:

```diff
-	cmpFn   string
+	CmpFn   string
```

`template.Template` cannot access unexported fields. When you call `{{.cmpFn}}`, it will fail to find the func

I also added an extra fields `Kinds` to enable the template to loop thru.

```diff
+       {{$numargs := .NumArgs}}
+       {{$fn := .CmpFn}}  
-        x := val1.({{asType .}}))
-        {{if ge .NumArgs 2 -}} y := val2.({{asType .}})){{end}}
-        {{if ge .NumArgs 3 -}} z := val3.({{asType .}})){{end}}
+        x := val1.({{asType .}})
+        {{if ge $numargs 2 -}} y := val2.({{asType .}})){{end}}
+        {{if ge $numargs 3 -}} z := val3.({{asType .}})){{end}}
```

When you're inside a `{{range .ANYTHING}}`, you basically enter a new scope (you technically enter the scope of `[]reflect.Kinds`. ``{{.NumArgs}}`  and `{{.CmpFn}}` will not be accessible once in the loop. Threrefore you have to store them in variables `{{$numargs := .NumArgs}}`

chewxy:
p/s I think by now you'd have notice my naming scheme for things are pretty shit... please feel free to rename anything/variables to something that makes more sense to you

kabaka0:
Awesome, thanks for the clarification. Getting it to work feels like magic!

chewxy:
Out of curiosity, why is the `destroyMultIterator` being called? If it's being called by `runtime.Finalize` you wouldn't need to manually call it - it'll be called by the Go GC

chewxy:
Additionally your comments will cause GoVet to scream:

Should change from this:

```go
// {{.Name}}(val) sets the mask to true where the corresponding data is {{.Desc}} val.append.append
// val must be the same type as the tensor
```

to 

```go
// {{.Name}} sets the mask to true where the corresponding data is {{.Desc}} val.append.append
// val must be the same type as the tensor
```

kabaka0:
> Out of curiosity, why is the destroyMultIterator being called?

The  code is me just learning how to use genlib, neglect its content. 

That said though, now that I think about it,  GC is not the best time to do it. If you look at the code, it simply returns a couple of slices back to the pool. It does not really destroy the Multiterator struct itself - I guess I named it incorrectly.

```
// destroyMultIterator creates a new MultIterator from a list of dense tensors
func destroyMultIterator(it *MultIterator) {
	if cap(it.masks) > 0 {
		ReturnMaskList(it.masks)
		it.masks = nil
	}
	if cap(it.trackIdx) > 0 {
		ReturnInts(it.trackIdx)
		it.trackIdx = nil
	}
}
```

chewxy:
I think it's aptly named though. I also think what it does is the correct thing. I just think that it should be a thing that the GC does (see: https://golang.org/pkg/runtime/#SetFinalizer)

kabaka0:
Aah, makes sense, especially given the fact that most iterators never leave the scope of a single function. Will change that.

kabaka0:
Some more stuff. Amongst several changes, added routines to make the masked array functionality mimic numpy as much as possible ( many of the function names are camel case versions of the equivalent numpy function). As promised, added support for masked arrays to the io and format routines. 

Trying really hard to make sure that I don't lower your stellar test coverage and so have added tests for almost everything (other than the masked version of apply which I have only just incorporated - will be added in due course).

kabaka0:
For some reason I thought it was 90% - I mistook the target for the current figure!

chewxy:
FYI you should merge latest changes from master - I fixed a bug with regards to the `overlap()` method

chewxy:
Hey @kabaka0 

wanted to check in with you to see how you're going with this - the #93 PR is getting ready quite soon. Want to see if you can merge first or I merge first - it's gonna affect the `*Dense` a bit - new flags that indicate that the memory is manually managed for example

chewxy:
Additionally, I'd like to package this PR, #93 and #7 and #8 into a release tag (let's call it 0.5).

kabaka0:
I've been overloaded with work recently so pace slowed down, but I should
get back to finishing it over the next few days. I made some further
rewrites to remove some features that caused code bloat and yet will only
have niche application. It's ok if you push changes, I will make sure to
merge them. But I am getting back onto it.

chewxy:
Ignore my last comment. A colleague reminded me that LSTMs and dynamic networks do not run on GPU yet.  That means I still have some ways to go to adapt `*LispMachine` into using CUDA. So the merge is some ways away

kabaka0:
Noted, thanks for the heads up, and apologies for the delayed response on
my end.

kabaka0:
@chewxy  If CI passes then this should be good to merge. Initially, I had designed the mask to potentially be stored in a smaller slice than the data (allowing a mask element to apply to an entire data dimension, as an example). I've scrapped that for now, I felt that it made the codebase grow significantly while reducing performance, because it required plenty of if-statements inside of loops impacting performance, and also required quite a bit of wrangling while implementing operations like stack. I've switched back to the simpler way of the mask having the same size as the data, and added support for masked to the various matrix operations, as well as most functions in genlib. I've tried to write tests for all changes, if any gaps we can plug them as we go along.

Sorry for the delay in getting to this point, I was not able to dedicate as much time to this over the last few weeks as I had hoped due to other commitments.

chewxy:
Can you add yourself to the list of SIGNIFICANT CONTRIBUTORs. I'll check the code later in a bit

kabaka0:
Sure, will do that now. Getting a fatal cuda.h error during CI, thought I had merged your recent Travis updates but it does not seem like that worked.

chewxy:
Nah you have it right. The error is this:

```
Failed to fetch http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1404/x86_64/./cuda-visual-tools-8-0_8.0.61-1_amd64.deb  Hash Sum mismatch
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

nvidia keeps updating their packages I think

kabaka0:
Forgot to add - reverted AP back to your original design given that the mask now shares same strides as data. Also added Iterator interface, just to make it easier to switch between iterator types without modifying code.

chewxy:
Posted a query on nvidia's devtalk. What I'll do later this evening if nvidia hasn't replied is I'll do a manual download and check and merge - I've got spare machines somewhere...

tensor/.gitgnore
```diff
@@ -0,0 +1,2 @@
+*.vscode
```
chewxy: I believe it should be `*.vscode/*`

tensor/.gitgnore~
```diff
@@ -0,0 +1 @@
+*.vscode
```
chewxy: this gedit cache file should be removed

tensor/dense.go
```diff
@@ -57,7 +63,7 @@ func recycledDenseNoFix(dt Dtype, shape Shape, opts ...ConsOpt) (retVal *Dense)
 	return
 }
 
-func newDense(dt Dtype, size int) *Dense {
+func newDense(dt Dtype, size int, maskSize ...int) *Dense {
```
chewxy: `maskSize` doesn't seem to be used?

Probably should be removed

tensor/dense.go
```diff
@@ -267,3 +342,178 @@ func (t *Dense) shallowClone() *Dense {
...
+}
+
+// SoftenMask forces the mask to soft
+func (t *Dense) SoftenMask() bool {
+	t.maskIsSoft = true
+	return t.maskIsSoft
+}
+
+// MaskFromSlice makes mask from supplied slice
+func (t *Dense) MaskFromSlice(x interface{}) {
```
chewxy: This is more a curiosity than anything - is there any reason why there are so many construction options for `MaskFromSlice` ? Wouldn't a `[]bool` suffice?

tensor/dense_io.go
```diff
@@ -71,8 +73,21 @@ func (t *Dense) WriteNpy(w io.Writer) (err error) {
 	bw.Write([]byte(header))
 
 	bw.seq = 0
-	for i := 0; i < t.len(); i++ {
-		bw.w(t.Get(i))
+	if t.IsMasked() {
+		fillval := t.FillValue()
+		it := FlatMaskedIteratorFromDense(t)
+		runtime.SetFinalizer(it, destroyIterator)
```
chewxy: `runtime.SetFinalizer` should be set in `FlatMaskedIteratorFromDense`

See comment in that function.

tensor/iterator.go
```diff
@@ -172,6 +368,73 @@ func (it *FlatIterator) Chan() (retVal chan int) {
...
+	it.FlatIterator = NewFlatIterator(ap)
+	it.mask = mask
+	return it
+}
+
+// FlatMaskedIteratorFromDense creates a new FlatMaskedIterator from dense tensor
+func FlatMaskedIteratorFromDense(tt *Dense) *FlatMaskedIterator {
+	it := new(FlatMaskedIterator)
+	it.FlatIterator = FlatIteratorFromDense(tt)
+	it.mask = tt.mask
```
chewxy: add `runtime.SetFinalizer` here instead.

If you set the finalizer here, you don't need to set the finalizer everywhere else. When the iterator falls out of reference (no other object holds a reference to the iterator anymore, and the function in which the iterator was created has fallen out of scope), the finalizer `destroyIterator` will be called. 

Let the runtime handle this, instead of manually managing it - I can't seem to find anywhere where you create an iterator and don't destroy it.

tensor/dense_io.go
```diff
@@ -95,15 +111,25 @@ func (t *Dense) WriteCSV(w io.Writer, formats ...string) (err error) {
 	}
 
 	cw := csv.NewWriter(w)
-	it := NewFlatIterator(t.AP)
+	it := IteratorFromDense(t)
+	runtime.SetFinalizer(it, destroyIterator)
```
chewxy: remove this.

tensor/dense_io_test.go
```diff
@@ -27,7 +27,7 @@ func TestSaveLoadNumpy(t *testing.T) {
 
 	script := "import numpy as np\nx = np.load('test.npy')\nprint(x)"
 
-	cmd := exec.Command("python")
+	cmd := exec.Command("python2")
```
chewxy: I assume the masked `Dense` is binary compatible with npy?

NICE :+1:

tensor/dense_mask_filling.go
```diff
@@ -0,0 +1,125 @@
...
+		for i := range sliceList {
+			tt, err := tc.Slice(nil, sliceList[i])
+			if err != nil {
+				ts := tt.(*Dense)
+				ts.Memset(fillval)
+			}
+		}
+	default:
+		it := IteratorFromDense(tc)
+		runtime.SetFinalizer(it, destroyIterator)
```
chewxy: remove

tensor/dense_mask_filling.go
```diff
@@ -0,0 +1,125 @@
...
+		for i := range sliceList {
+			tt, err := t.Slice(nil, sliceList[i])
+			if err != nil {
+				ts := tt.(*Dense)
+				ts.Memset(fillval)
+			}
+		}
+	default:
+		it := IteratorFromDense(t)
+		runtime.SetFinalizer(it, destroyIterator)
```
chewxy: remove

tensor/dense_mask_inspection.go
```diff
@@ -0,0 +1,278 @@
...
+		m := t.mask
+		if len(t.mask) == t.Size() {
+			for _, v := range m {
+				if !v {
+					return false
+				}
+			}
+		} else {
+			it := IteratorFromDense(t)
+			runtime.SetFinalizer(it, destroyIterator)
```
chewxy: remove

tensor/dense_mask_inspection.go
```diff
@@ -0,0 +1,278 @@
...
+		m := t.mask
+		if len(t.mask) == t.Size() {
+			for _, v := range m {
+				if v {
+					return true
+				}
+			}
+		} else {
+			it := IteratorFromDense(t)
+			runtime.SetFinalizer(it, destroyIterator)
```
chewxy: remove

tensor/dense_mask_inspection.go
```diff
@@ -0,0 +1,278 @@
...
+		m := t.mask
+		if len(t.mask) == t.Size() {
+			for _, v := range m {
+				if v {
+					count++
+				}
+			}
+		} else {
+			it := IteratorFromDense(t)
+			runtime.SetFinalizer(it, destroyIterator)
```
chewxy: remove

tensor/dense_mask_inspection.go
```diff
@@ -0,0 +1,278 @@
...
+----------*/
+
+// FlatNotMaskedContiguous is used to find contiguous unmasked data in a masked array.
+// Applies to a flattened version of the array.
+// Returns:A sorted sequence of slices (start index, end index).
+func (t *Dense) FlatNotMaskedContiguous() []Slice {
+	sliceList := make([]Slice, 0, 4)
+
+	it := IteratorFromDense(t)
+	runtime.SetFinalizer(it, destroyIterator)
```
chewxy: Remove

tensor/dense_mask_inspection.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+// FlatMaskedContiguous is used to find contiguous masked data in a masked array.
+// Applies to a flattened version of the array.
+// Returns:A sorted sequence of slices (start index, end index).
+func (t *Dense) FlatMaskedContiguous() []Slice {
+	sliceList := make([]Slice, 0, 4)
+
+	it := IteratorFromDense(t)
+	runtime.SetFinalizer(it, destroyIterator)
```
chewxy: remove

tensor/dense_mask_inspection.go
```diff
@@ -0,0 +1,278 @@
...
+// Applies to a flattened version of the array.
+// Returns: A pair of ints. -1 if all values are masked.
+func (t *Dense) FlatNotMaskedEdges() (int, int) {
+	if !t.IsMasked() {
+		return 0, t.Size() - 1
+	}
+
+	var start, end int
+	it := IteratorFromDense(t)
+	runtime.SetFinalizer(it, destroyIterator)
```
chewxy: remove

tensor/dense_mask_inspection.go
```diff
@@ -0,0 +1,278 @@
...
+// FlatMaskedEdges is used to find the indices of the first and last masked values
+// Applies to a flattened version of the array.
+// Returns: A pair of ints. -1 if all values are unmasked.
+func (t *Dense) FlatMaskedEdges() (int, int) {
+	if !t.IsMasked() {
+		return 0, t.Size() - 1
+	}
+	var start, end int
+	it := IteratorFromDense(t)
+	runtime.SetFinalizer(it, destroyIterator)
```
chewxy: remove

tensor/genlib/dense_argmethods.go
```diff
@@ -46,15 +45,18 @@ func (t *Dense) {{.ArgName}}(axis int)(retVal *Dense, err error){
 		newAP = t.AP.Clone()
 	}
 	defer ReturnAP(newAP)
-
-	it := NewFlatIterator(newAP)
+	
+	it :=IteratorFromDense(t)
+	runtime.SetFinalizer(it, destroyIterator)
```
chewxy: remove

chewxy:
I've not actually done a full thorough review. Whenever I add a new comment, github's js uses too much resources and causes my browser to crash, so I've temporarily ended the review

tensor/.gitgnore
kabaka0: Done

tensor/.gitgnore~
kabaka0: Done

tensor/dense.go
kabaka0: Done. Legacy from previous implementation.

tensor/dense.go
kabaka0: I had some applications where the mask needed to be set based on the output of another function that returned floats - this was simply to skip writing code to convert these arrays to Bools beforehand.

tensor/dense_io.go
kabaka0: Done

tensor/dense_io.go
kabaka0: Done

tensor/iterator.go
kabaka0: Done

tensor/dense_io_test.go
kabaka0: Should be. As far as I could tell this was the most efficient  and universal way to do it - simply setting masked values to the fill value means that arrays can be opened in/from numpy. If you want to save the raw mask too, then I figured we could just leave that to the programmer to implement in whatever format they prefer.

tensor/dense_mask_filling.go
kabaka0: Done

tensor/dense_mask_inspection.go
kabaka0: Done

tensor/dense_mask_inspection.go
kabaka0: Done

tensor/dense_mask_inspection.go
kabaka0: Done

tensor/dense_mask_inspection.go
kabaka0: Done

tensor/dense_mask_inspection.go
kabaka0: Done

tensor/dense_mask_inspection.go
kabaka0: Done

tensor/dense_mask_inspection.go
kabaka0: Done

tensor/genlib/dense_argmethods.go
kabaka0: Done

tensor/dense.go
chewxy: Ah. That makes sense

kabaka0:
There are a couple of  missing bits - I need to ensure that reduction is covered, after which genlib changes should be complete. It slipped past me.

kabaka0:
Just remembered. I ported gsl's randist library to pure Go for my own use.
It might be a neat addition to tensor, and close another of the gaps to
numpy, adding a few dozen random distributions and special functions. I can
create a pull request if you are interested. A problem though is license
compatibility - gsl is GPLv3, while Gorgonia is Apache 2. The code would
count as a derived work, and so would still require a GPL license. If you'd
like the functionality and can figure out a way around that, I can get that
incorporated into tensor.

chewxy:
GPL isn't compatible with Apache 2 unfortunately. :S

chewxy:
Sweet! This thing passed. Will merge after gym

kabaka0:
Cool! :thumbsup:

PR #122, opened 2017-06-17
Refactor mask

chewxy:
refactored the masked tensor construction - removed the need for (...[]bool), instead, added an additional `addMask()` method.

### DIFF
--- a/tensor/dense_io.go
+++ b/tensor/dense_io.go
@@ -471,92 +471,93 @@ func (t *Dense) GobDecode(p []byte) (err error) {
 		if err = decoder.Decode(&data); err != nil {
 			return
 		}
-		t.fromSlice(data, mask)
+		t.fromSlice(data)
 	case reflect.Int8:
 		var data []int8
 		if err = decoder.Decode(&data); err != nil {
 			return
 		}
-		t.fromSlice(data, mask)
+		t.fromSlice(data)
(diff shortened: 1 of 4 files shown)


tensor/dense_io.go
```diff
@@ -471,92 +471,93 @@ func (t *Dense) GobDecode(p []byte) (err error) {
...
+		t.fromSlice(data)
 	case reflect.String:
 		var data []string
 		if err = decoder.Decode(&data); err != nil {
 			return
 		}
-		t.fromSlice(data, mask)
+		t.fromSlice(data)
 	}
+	t.addMask(mask)
```
chewxy: This is where all the addMasks are replaced @kabaka0

chewxy:
I'd love your thoughts on this @kabaka0 . So far I have ran all the tests and it all passed. Is there a particular reason why you added a `(...[]bool)` option for `fromSlice`?

tensor/dense_io.go
kabaka0: This looks fine.

kabaka0:
That particular construction was probably a legacy of some code I had previously written before submitting the pull request. I agree that having a separate addMask() method probably makes it less confusing for users that never have reason to use the mask.

chewxy:
Cheers mate.

PR #136, opened 2017-08-17
Debugcheck from egonelbre

chewxy:


chewxy:
Closed. Issue was averted alltogether

Issue #116, opened 2017-05-14
Adopt "dep" as the official installation mechanism?


NDari:
With the possible integration of the `dep` package manager into the Go toolchain, it may be worth it to adopt it as the official installation method for Gorgonia. There are a good bit of packages to install to fully utilize Gorgonia, which may scare people new to Go and/or programming in general.

- [] provide a Gopkg.toml with explicit versions of the libraries whenever possible
- [] provide a Gopkg.lock 
- [] Add the vendor directory to the .gitignore file so that it is not checked in.
- [] Provide an installation section in the readme showing how to use `dep` to install Gorgonia and how to test if your installation was successful.

damienstanton:
This makes sense, `dep` is simple enough and can help people who might not be familiar with Go vendoring quirks.

NDari:
I think I should clarify that this is for developers. `go get` does the trick for people who just want to use the package. I think we need a uniform way to a developer to get started.

rereading my issue I wasnt sure that I made that clear.

chewxy:
I fully agree

NDari:
@chewxy The work here is mainly done, however there is two issues:

- In regaurd to wanting explicit versions of the packages used, this is a WIP. Part of this work would be greatly helped if you (or I...) can version your libraries which are used by Gorgonia. For instance 

[[dependencies]]
  branch = "master"
  name = "github.com/chewxy/math32"

[[dependencies]]
  branch = "master"
  name = "github.com/chewxy/vecf32"

I would much rather have an explicit release of these (x.x.x) So that we can lock things down and are able to develop these packages independantly without the risk of breaking anything accidentally. 

I am opening some issues on the other packages which we rely on, asking if they can add some versioned releases.

- The second issue is the URL of your packages. Which ones do you plan to move under Gorgonia? for instance, the two packages above are to remain personal, or are they also getting moved?

chewxy:
Oh right, package `cu`, `vecfxx` will also all be moved to Gorgonia

NDari:
Here is the complete list:

```
[[dependencies]]
  name = "github.com/awalterschulze/gographviz"
  version = "~2.0.0"

[[dependencies]]
  branch = "master"
  name = "github.com/chewxy/cu"

[[dependencies]]
  branch = "master"
  name = "github.com/chewxy/gorgonia"

[[dependencies]]
  branch = "master"
  name = "github.com/chewxy/hm"

[[dependencies]]
  branch = "master"
[...]

PR #139, opened 2017-08-20
Dep

chewxy:
see also #116


NDari:
I am second guessing myself now... should this have been merged?

chewxy:
que pasa? I gave you admin rights for good reason.

NDari:
I wondered if you wanted to address the final issue with the documentation in this PR or not. We can alway address that later!

chewxy:
Next PR is good.. I have no idea how to document `dep` usage yet other than
`dep init` and `dep ensure`

Plus several things need to be pulled in before dep can probably be used
properly

PR #168, opened 2017-10-05
Tensordot

siquus:
My hello world on github contributions. Basically, I would just like to know if it makes sense to pursue this direction (and add more to it) and maybe receive some pointers for improvement.

As I am new to working on computational graphs, if you happen to have any tips on literature that would be very helpful.

Thanks!


tensor/api_arith.go
```diff
@@ -459,3 +459,20 @@ func Outer(a, b Tensor, opts ...FuncOpt) (retVal Tensor, err error) {
 	}
 	panic("Unreachable")
 }
+
+// Contract performs a contraction of given tensors along given axes
+func Contract(aAxes, bAxes []int, a, b Tensor) (retVal Tensor, err error) {
```
chewxy: can the API of this change to 

`func Contract(a, b Tensor, axesA, axesB []int) (retVal Tensor, err error)`?

siquus:
Alright, Bitbucket would have merged this itself ;-)

I took the argument order for contract from operations.go/concat. But yes, it sure is more intuitive to have the tensors come first.

chewxy:
Thank you once again

PR #173, opened 2017-11-10
Tensordot ad op interface

siquus:


chewxy:
there is some issue with Go-tip. I'll merge immediately once that's fixed

siquus:
Golang changed something about fmt flag handling...it doesn't like the "%#-d " hack in formatter any longer..

chewxy:
Fixed the bugs in tip. Will be merging #174 then this

chewxy:
So there's an actual issue with `go vet`. https://github.com/golang/go/issues/22701#issuecomment-344146948

PR #174, opened 2017-11-13
Fixed up the issues that gotip has found

chewxy:
Go is a lot more religious in checking for string interpolation and formatting now.


chewxy:
Fuuuuuuuuuuuuuuuuuuuuuuuuck 
https://github.com/golang/go/issues/22701

PR #176, opened 2017-11-19
Tensordot ad op sd op interface

siquus:
I am aware that TestMul now fails, but I can't see why: Do you in general agree with the changes to Mul?


chewxy:
Checking

siquus:
Set up testing on my IDE and then my stupid errors weren't that hard to find ;-)

chewxy:
Good to merge? (tell me when ready)

chewxy:
@siquus I want to merge this into v0.8.0. As a result, I've ported over your changes to gorgonia/tensor and I'm awaiting your greenlight for being OK to merge

siquus:
@chewxy  Should be fine to merge.

PR #179, opened 2017-12-17
Fix table formatting

bradleyjkemp:


chewxy:
Thank you. I'll merge this once the status passes.

PR #193, opened 2018-01-28
Tantanchen accuracy in values_untils.go

tantanchen:
There's something I don't understand about gorgonia. The Value interface has Data() which returns an interface{}. I did some research online and I think I understand what and how interfaces work in Go. But I don't actually know what the "target" and "predicted" types are going to be in this function.

From the logisticregression example, the `pred` node is a tensor of Boolean values. But it doesn't look like the node or the tensor has a Data(). So I don't think it "implements" Value{}. Intuitively, I feel that it must, so I think I'm just missing something. (I know what I did won't work, as you cannot index through an interface{})

Side note, after I figure out how to access the data in target and predicted, I'm going to need to check for different types. The Accuracy func should also work for categorical variables as well.


chewxy:
`Data()` returns an `interface{}`. That `interface{}` value depends on what the tensor represents. If the tensor represents a array of `float64` then you can cast it like that: `Data().([]float64)`. Or if it's float32 then `Data().([]float32)`

values_utils.go
```diff
@@ -1,247 +1,266 @@
...
+	switch vv := v.(type) {
+	case *dualValue:
+		setEngine(vv.Value, e)
+		setEngine(vv.d, e)
+	case tensor.Tensor:
+		tensor.WithEngine(e)(vv)
+	}
+}
+
+func Accuracy(target, predicted Value) (float64, error){
```
chewxy: Accuracy shouldn't be in the Gorgonia package at all. It should only be in the logistic regression package. Different machine learning projects would have different definitions of accuracy.

chewxy:
Also if you don't mind, perhaps you could set your CRLF settings in your git? https://help.github.com/articles/dealing-with-line-endings/#platform-all

chewxy:
Closing this since #191 has been merged

Issue #97, opened 2017-03-18
HM-like shape inference?


chewxy:
With reference to #96 , maybe it's time to create a HM-like system for shape inference, as per the package `hm`.  The current shape inference system is kinda crappy.

## Reference Material ##

Look at TensorFlow's shape inference.

Issue #96, opened 2017-03-18
Use negative shape dimensions for dynamic/variable shapes


chewxy:
Allow for such shapes: `(-1, 200)`. This allows for shapes that change at run time. 

## Pros ##

The running example used is in the context of generating logical forms coupled with a reinforcement learning algorithm.

In a logical form generator, a seq2seq network (or even just a maximum likelihood model paired with a traditional chart parser) would generate `n` logical forms, which is then scored, and weighted with the rewards. A pseudomodel is something like this:

```go
// elided: a complicated model that generates a rewards vector with dynamic size

weights := 500
w := NewVector(g, Float64, WithShape(weights), WithName("weights"), WithInit(GlorotU(1)))
x := NewMatrix(g, Float64, WithShape(weights, rewards), WithName("xs"))
r := NewVector(g, Float64, WithShape(rewards), WithName("reward weighting"))
a := Must(Tanh(Must(Mul(w, f)))
WithName("activated")(a)
rWeighted := Must(HadamardProd(r, a))
```

Here `rewards` is the size of dynamically sized vector that will change size based on whatever the reinforcement learning algorithm (or a seq2seq network) returns. 

Symbolically, we can work out the shape of `a` to be `(rewards)`. Allowing negative shape dimensions would allow for variable shapes. 

## Cons ##
*  Huge footprint - this change affects many moving parts, especially with allocating the slices.
* Potential performance issues - currently `Value` stored in the `*Node` doesn't change once it's been allocated. Additional values are merely copied into it. If the shape changes, then plenty more allocations and freeing would be required.

## Current Workarounds ##
Currently, a manual update to the shapes are required to each node at each runtime. While in the example above it's very easy (as there are only two `*Nodes` to update), especially with a `LispMachine`, This gets very complicated as a new shape calculator needs to be created with each application.

### Alternative Slutions ###
* Create `AsShape` `Op`s. - this sidesteps all the 

### Handling Negative Shape Dimensions ###
* Treat the negative dimensions as a size hint/signal.
* `TotalSize` is calculated the same way, but with an additional `abs()` called to give a total positive size.

## TODO ##
* [ ] Gather all the places this change would affect
* [x] Propose a mechanism for handling negative shape dimensions.
* [x] Propose alternatives (with pros and cons).

Issue #91, opened 2017-02-22
Better Package Organization


chewxy:
`Op` can be put into its own package
`Node` and `Graph` can be put into its own package
`Compile` can be put into its own package
`VM` can be put into its own package

chewxy:
Experimental only. Culmination of rethinking that happened around #65

pbberlin:
There is no tool to "break out" a type into a subpackage.
Alex Donovan started a tool "sock drawer" but it seems unfinished.
Once you have subpackages, you can move them around easily, using
`gomvpkg -from github.com/a/b/c/d  -to github.com/a/d`

If you encounter circular dependencies, after subpackaging,
you need to create further packages with the "joined" types.

Still I enjoyed the results of subpackaging my own project.

Regards 
  Peter

chewxy:
Yeah... I was doing it by hand. There are a number of circular dependencies, and I am currently thinking about what the best move forwards is

owulveryck:
Part of this issue is implemented with #307

ArturTheAmazing:
Created a PR #481

PR #195, opened 2018-03-11
V0.9.0 

chewxy:
Ongoing notes:

* **CUDA**: Better CUDA support (IN PROGRESS)
    * ~ColMajor used by default if engine is CUDA.~ (ColMajor is supported, but defaults to using RowMajor for all the major cuBLAS versions. Careful reasoning of the parameters obviates the need for ColMajor by  default, which causes more headaches. It is still supported)
    * Transposition will be automatically done when performing transports back to CPU.
    * cudnn operations supported (IN PROGRESS) (note: these are the ones I use more often hence gets bigger attention):
        * [x] Conv2d
        * [x] Dropout
        * [x] Maxpool2d 
        * [x] BatchNorm
        * [x] Rectify
    * Other CUDA related optimizations
        *  [x] full cuBLAS support
* **New Ops**:
    * BatchNorm 
    * InvSqrt
    * CUDA enabled ops in `ops/nn` (preview for how things will start to look in v0.10.0)
* **New Features**:
    * Limited shape inference.  Working towards a calculus for shapes (first raised in #96 and #97).
* **Optimizations**:
    * Optimizations of basic ops to use engine functions if available, otherwise, fall back to using `Apply`, which adds a penalty from repeatedly calling functions.
    * Faster VMs (1 of 2 VMs): ~greedy goroutines grabs gigs from a priority queue. This causes faster execution of  code in general.~ (this is moved to a future version of 0.9.xx):
```
benchmark                           old ns/op      new ns/op      delta
BenchmarkTapeMachineExecution-8     3129074510     2695304022     -13.86%

benchmark                           old allocs     new allocs     delta
BenchmarkTapeMachineExecution-8     25745          25122          -2.42%

benchmark                           old bytes      new bytes      delta
BenchmarkTapeMachineExecution-8     4804578705     4803784111     -0.02%
````
* **Code generation**: some exported API is now auto generated
* **New Solver** : @ynqa added the Momentum solver.
* **Breaking API**: `Solver` now take a slice of `ValueGrad` instead of `Nodes`. `ValueGrad` is an interface, of which a `*Node` fulfils. An additional utility function `NodesToValueGrads` has been added to aid with refactoring.  This was done for two reasons:
    * ~The support for BatchNorm operation, which is a verily impure and highly stateful function. The BatchNorm Op has internal states that need to have their gradients updated as well. But the internal state of BatchNorm isn't really part of  the expression graph, and really it shouldn't be.~ Turns out there was a better API for `BatchNorm`. 
    * In the next version,  v0.10.0. We aim to  do [better package organization](https://github.com/gorgonia/gorgonia/issues/91) for managability.  With this API breaking change, the solver now is less dependent on the other parts of Gorgonia and can be easily separated.
* **Breaking Semantics**: A `gorgonia.VM` now implements `io.Closer`. It should be treated as a resource as well as a computation device - the VM must be `Close()`d in order for the resources acquired by the VM to actually be released. Turns out, automatic resource management is too difficult. Who'd thunk that?


docmerlin:
Awesome.  Excited about the col-major, stuff for more speed.

PR #199, opened 2018-04-07
Concurrent tape machine

chewxy:
```
benchmark                           old ns/op      new ns/op      delta
BenchmarkTapeMachineExecution-8     2871121602     2576590159     -10.26%

benchmark                           old allocs     new allocs     delta
BenchmarkTapeMachineExecution-8     25239          16661          -33.99%

benchmark                           old bytes      new bytes      delta
BenchmarkTapeMachineExecution-8     4804582632     4804524502     -0.00%
```


chewxy:
No longer necessary. See #224

PR #202, opened 2018-05-10
Сorrected typos, fixed ineffassign, gofmt, go_vet

trigun117:


solvers.go
```diff
@@ -1160,13 +1160,13 @@ func (s *BarzilaiBorweinSolver) Step(model Nodes) error {
 					contractionAxes[axis] = axis
 				}
 
-				valGradDiffscalarProd, err := tensor.Contract(valueDiff, gradDiff, contractionAxes, contractionAxes)
+				valGradDiffscalarProd, _ := tensor.Contract(valueDiff, gradDiff, contractionAxes, contractionAxes)
```
chewxy: This is not a good idea. @siquus how should errrors be handled for Barzilai Borwein. I'm going to guess the default is to return the error?

chewxy:
Thanks for this. More importantly thanks for finding a potential bug

I have left a comment waiting @siquus' reply on how to handle errors. If he doesn't respond in 12 hrs, can you please fix it such that the err is not shadowed?

trigun117:
@chewxy i will check what i can do if he not respond

chewxy:
Thanks :+1:  I'll merge this when ready.

siquus:
@trigun117 nice catch ...somebody... forgot to actually react to err ;-)
Failing silently at that point will be very hard to debug, so if you could please add a `return errors.new` or so along the lines "operationError, Contracting value / gradient difference" that would be nicer.
Thanks!

trigun117:
@chewxy updated code

chewxy:
LGTM. Merged

PR #210, opened 2018-06-03
Add momentum sgd algorithm

ynqa:
Add momentum sgd algorithm.

### DIFF
--- a/solvers.go
+++ b/solvers.go
@@ -73,7 +73,7 @@ func WithL2Reg(l2reg float64) SolverOpt {
 		case *VanillaSolver:
 			st.l2reg = l2reg
 			st.useL2Reg = true
-		case *MomentumSGDSolver:
+		case *Momentum:
 			st.l2reg = l2reg
 			st.useL2Reg = true
 		}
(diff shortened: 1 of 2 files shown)


solvers.go
```diff
@@ -898,6 +919,232 @@ func (s *VanillaSolver) Step(model []ValueGrad) (err error) {
...
+	l2reg    float64 // l2 regularization parameter
+	batch    float64 // batch size
+
+	useClip, useL1Reg, useL2Reg bool
+
+	cache []*dualValue
+}
+
+// NewMomentumSGDSolver creates a new MomentumSGDSolver with sane-ish default values
+func NewMomentumSGDSolver(opts ...SolverOpt) *MomentumSGDSolver {
```
chewxy: Can we call this `NewMomentum` instead?

solvers.go
```diff
@@ -898,6 +919,232 @@ func (s *VanillaSolver) Step(model []ValueGrad) (err error) {
 	return
 }
 
+// MomentumSGDSolver is the stochastic gradient descent optimizer with momentum item.
+type MomentumSGDSolver struct {
```
chewxy: `type Momentum struct`

Because in v0.10.0 all the solvers will be moved to their own package.. anything new should follow the new scheme

Issue #214, opened 2018-06-29
dep ensure shows a warning


ifraixedes:
`dep ensure` shows the following warning

```
Warning: the following project(s) have [[constraint]] stanzas in Gopkg.toml:

  ✗  github.com/chewxy/vecf64
  ✗  golang.org/x/tools
  ✗  gorgonia.org/vecf32

However, these projects are not direct dependencies of the current project:
they are not imported in any .go files, nor are they in the 'required' list in
Gopkg.toml. Dep only applies [[constraint]] rules to direct dependencies, so
these rules will have no effect.

Either import/require packages from these projects so that they become direct
dependencies, or convert each [[constraint]] to an [[override]] to enforce rules
on these projects, if they happen to be transitive dependencies.
```

Should those dependencies converted as suggested or not?

chewxy:
Quite sure those are definitely imported. But yes they should be converted

chewxy:
Thanks for the fix in #215 @ifraixedes

Issue #213, opened 2018-06-29
tests are not compiling 


ifraixedes:
I've just cloned the project and ran `dep ensure`.

I wanted to run the tests but when I tried, I got

```
# gorgonia.org/gorgonia
./walker.go:43:43: cannot use g (type *ExprGraph) as type graph.Directed in argument to topo.SortStabilized:
        *ExprGraph does not implement graph.Directed (wrong type for Edge method)
                have Edge(int64, int64) graph.Edge
                want Edge(graph.Node, graph.Node) graph.Edge
./walker.go:53:33: cannot use g (type *ExprGraph) as type graph.Directed in argument to topo.Sort:
        *ExprGraph does not implement graph.Directed (wrong type for Edge method)
                have Edge(int64, int64) graph.Edge
                want Edge(graph.Node, graph.Node) graph.Edge
./differentiation_test.go:18:26: cannot use g (type *ExprGraph) as type graph.Directed in argument to topo.Sort:
        *ExprGraph does not implement graph.Directed (wrong type for Edge method)
                have Edge(int64, int64) graph.Edge
                want Edge(graph.Node, graph.Node) graph.Edge
./differentiation_test.go:44:26: cannot use g (type *ExprGraph) as type graph.Directed in argument to topo.Sort:
        *ExprGraph does not implement graph.Directed (wrong type for Edge method)
                have Edge(int64, int64) graph.Edge
                want Edge(graph.Node, graph.Node) graph.Edge
./graph_test.go:92:34: cannot use g (type *ExprGraph) as type graph.Directed in argument to topo.Sort:
        *ExprGraph does not implement graph.Directed (wrong type for Edge method)
                have Edge(int64, int64) graph.Edge
                want Edge(graph.Node, graph.Node) graph.Edge
FAIL    gorgonia.org/gorgonia [build failed]
```

I'm wondering what I missed.

Thanks.

chewxy:
Nothing. The dep files are outdated. Gonum has since upgraded. If you have the latest Gonum, the package should compile.

The reason why dep files are not updated - vgo is coming.

PR #215, opened 2018-07-02
Remove waring of dep, update Gonum & minor fix in example

ifraixedes:
To wrap was more thorough commented in commit messages:

1. Remove warning in deps https://github.com/gorgonia/gorgonia/issues/214
2. Updated Gonum in Gopkg ([why not despite of _vgo_ is coming?[(https://github.com/gorgonia/gorgonia/issues/213#issuecomment-401522088))
3. Minor fix in one of the examples

I believe that _2_, is good, although _vgo_ is coming, however if you prefer not to and you want to keep the rest of the commits, let me know and I will drop such commit.

I could make a comment about _1_, before making the change, however, because of the minor change, it was easier to make a commit rather than trying to explain by words on the issue; however, I know that you may prefer other thing, so don't hesitate to comment what to do or reject such commit.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -3,7 +3,14 @@
 
 [[projects]]
   name = "github.com/awalterschulze/gographviz"
-  packages = [".","ast","internal/errors","internal/lexer","internal/parser","internal/token"]
+  packages = [
+    ".",
+    "ast",
+    "internal/errors",
+    "internal/lexer",
+    "internal/parser",
+    "internal/token"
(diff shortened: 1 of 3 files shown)


chewxy:
Hi,

my response to that is more along the lines of "vgo is coming... dep will be lazily updated as I don't really see the point of chasing with dep any more". In the meantime I'm learning more about vgo and quirks, so... there's taht..

I'm gonna merge this after the tests pass.

PR #225, opened 2018-08-19
[Enhancement] Suppressed assembly dependency

chewxy:
* popcnt has removed and replaced by bits/OnesCount64
* clz has been removed and replaced by bits/LeadingZeros64
* divmod pure go implementation has been imported from the tensor
package

some build tags has been added for the compilation on a wasm backend


pointlander:
The wasm build tag is not needed. mathutils_go.go should be built when !amd64 or noasm. The code from https://github.com/gorgonia/tensor can be copied. This should allow all architectures (including wasm) to be supported.

owulveryck:
@chewxy, are you ok to merge this PR?

PR #240, opened 2018-09-30
[fix] panix if a constant already exists

owulveryck:
if a Constant already exists, retVal is not set and the defer function panic


chewxy:
Feel free to merge when it's green @owulveryck  (spotty internets)

owulveryck:
My mistake, Gorgonia is ok, I was just badly using it!

PR #242, opened 2018-10-01
[fix] Explicit error message is the graph has no input

owulveryck:
While testing the "Constant" operation into "gorgonnx", I have created
a graph only holding a constant node.
This is obviously not useful, but it leads to a panic with an "index out
of range".

This gives a more explicit message


chewxy:
Good to merge?

Issue #223, opened 2018-08-17
The Broadcast function is exported but not usable outside of the package


owulveryck:
I need to implement a "add" operator for two tensors with a broadcasting mechanism as described [here](https://github.com/onnx/onnx/blob/master/docs/Broadcasting.md).

The [`Broadcast`](https://godoc.org/gorgonia.org/gorgonia#Broadcast) function seems to be a perfect fit for this. Moreover, the [test](https://github.com/gorgonia/gorgonia/blob/master/broadcast_test.go#L56) is partially implementing what I am trying to do.
But nor the `ʘBinaryOperatorType` or any other `binOp` implementations are exported.

Therefore the Broadcast function can only be used within the Gorgonia package.

Maybe we should make it private to avoid confusion in the documentation and expose "Broadcasted version" of some operators instead? What do you think?

chewxy:
I had a look at `Broadcast` - hadn't touched it in years. That and the elementwise operations should be hung in the Shitty Software Engineering Hall of Fame. Will get in contact with you to work on this further.

chewxy:
https://github.com/gorgonia/gorgonia/releases/tag/v0.9.0-beta < -- this is out so your PR can be based off this

deasmhumnha:
Trying to implement a column-wise normalization and running into the same problem. I don't feel I have enough of an understanding of `gorgonia` yet to attempt a PR, but I am curious about the design choice of requiring an explicit broadcasting pattern rather than having an implicit set of rules that act on  ndarrays based on their shapes (as Numpy does).

chewxy:
I put it to you that having implicit broadcasting leads to subtle errors you may not have caught. Even when you are familiar with the rules.

The rule is quite simple: a lower dimensioned tensor is an embedding within the higher dimension tensor.  Easy to state, but not easy to remember when working on complicated things. Having an explicit broadcast meant that you know exactly where the broadcast happens in code.

As for not making it  usable (the function is exported, the op isn't)... that's really more of a technical issue. I believe at one point the circle with a dot was considered exported. Or I may be wrong.

deasmhumnha:
I figured that was the primary motivation. It's a code clarity vs. code complexity trade-off for sure. Also, as owulveryck mentioned, wouldn't you also have to export the various lowercase implementations as well? It would also be nice if you could broadcast a custom `BinaryOp`.

chewxy:
Send PR! :)  (you seem to be aware of the `BinaryOp` type)

[...]

PR #246, opened 2018-10-31
Proposal for a new API for broadcasting

owulveryck:
**WARNING** This PR is changing the API.

Regarding the [issue 223](https://github.com/gorgonia/gorgonia/issues/223) about broadcasting, I propose the following change:

* Making the `Broadcast` method private
* Changing the `binOps` implementations to take the `BroadcastPattern` as argument.

for example:

```go
 // Add perfors a pointwise add operation.
 // Broadcasting is applied as mentioned by BroadcastPattern.
 // Set it to 0 to avoid broadcasting
 func Add(a, b *Node, broadcastPattern BroadcastPattern) (*Node, error) {
         if broadcastPattern == 0 {
                 return binOpNode(newElemBinOp(addOpType, a, b), a, b)
         } else {
                 return broadcast(addOpType, a, b, broadcastPattern)
         }
 }
```

I have updated the api generator, adapted the existing code, tests and example.

I have though about a more transparent method by using optional parameters:

```go
func Add(a, b *Node, pattern ...byte) 
```

and forcing the pattern to be the left side of a Pattern as expected by the `NewBroadcastPattern` function, but that was less elegant.

What do you think?


chewxy:
Omfg I completely failed to reply to this. 

Thoughts:

Having these two classes of functions would be nicer:
```
type broadcastableBinOp func(a, b *Node, p BroadcastPattern) (*Node,error)
type binOp func(a, b *Node) (*Node, error)
```

This way we keep the minimum possible changes to the API. `Add` becomes `Add` and `BroadcastAdd`. The `Add` function signature doesn't change. This keeps backwards compatibility

owulveryck:
Another idea: dealing with optional arguments:

` type binOp func(a, b *Node, broadcastPattern ...byte) (*Node, error)`

It would not break the API if no broadcast is applied.  

`binOp(a,b)` would work as well as `binOp(a,b,NewBroadcastPattern([]byte{0,1},nil)...)` would also work.

It is not really idiomatic Go (passing an array would be more elegant), but it brings best of both worlds (API stability and new feature).

WDYT?

PR #269, opened 2019-02-24
Attempt to fix Max

pa-m:
Hi,
I modified opMax.InferShape, opMax.Do
and added TestMax
it's probably a bit naïve but it does work for me on CPU
Best regards


chewxy:
LGTM. Will merge by eod (going to work now, saw this too late)

PR #276, opened 2019-03-12
Update graph edge for gonum/graph API change

kortschak:
This will be necessary when gonum/gonum#867 is merged.


kortschak:
This is now necessary.

chewxy:
thanks

PR #281, opened 2019-04-14
use a sync.Map instead of a map for dimSizerPool to prevent race condition

owulveryck:
All of the tests in onnx can run in parallel. 

In this case, every subtest creates a new `ExprGraph`. The problem is that the `dimSizerPool` is not safe for concurrent use (as it is a map) and running several tests in parallel leads to a race condition.
This PR replaces the `map` by a `sync.Map`.

This patch removes the race condition but may have an impact on the performances.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,10 @@ branches:
   only:
     - master
 go:
-  - 1.8.x
   - 1.9.x
   - 1.10.x
   - 1.11.x
+  - 1.12.x
   - tip
 
 env:
--- a/perf.go
+++ b/perf.go
@@ -82,25 +82,30 @@ func returnValue(v Value) {
 	}
 }
 
-var dimSizerPool = make(map[int]*sync.Pool)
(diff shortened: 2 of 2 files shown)


owulveryck:
If you want to test:

```
go get github.com/owulveryck/onnx-go
git checkout bd6418dbc40f14e9812bbbdcf54e26910d541637
cd $GOPATH/src/github.com/owulveryck/onnx-go/backend/x/gorgonnx
go test -race
```

owulveryck:
It’s good to merge but it forces to use go >= 1.8. So i’d rather have a review from @chewxy

chewxy:
Will look tonight. Sorry just been quite busy

chewxy:
LGTM. Will run a suite of benchmarks to figure out the loss of perf using sync map, but otherwise, all good to merge (if I don't merge by tonight, you have the ability to merge by yourself @owulveryck )

PR #290, opened 2019-06-03
New operator: Leaky relu

owulveryck:


owulveryck:
INot the most efficient implementation I think, but if it's ok for you @chewxy this is good to merge.

chewxy:
Gimme about 8 hrs.. Will 👀 at it soon

PR #292, opened 2019-06-05
Not using the full path confuses `go mod` for internal `gorgonia.org/gorgonia` imports

### LINKED ISSUE #277: Cannot build with latest version of gonum and Go modules
Steps to reproduce:

```shell
rm -f go.mod && rm -f go.sum && rm -rf vendor
go mod init `basename $(pwd)` && go mod tidy; go mod vendor
go build -mod=vendor -o ./bin/foo ./cmd/service
```

Error received from build:

```go
# gorgonia.org/gorgonia
vendor/gorgonia.org/gorgonia/graph.go:569:2: cannot use e (type edge) as type graph.Edge in return argument:
	edge does not implement graph.Edge (missing ReversedEdge method)
vendor/gorgonia.org/gorgonia/node.go:756:15: cannot use e (type edge) as type graph.Edge in argument to n.g.SetEdge:
	edge does not implement graph.Edge (missing ReversedEdge method)
```

durp:
It erroneously decides that these must be external and creates a circular reference for any packages that attempt to `go get gorgonia.org/gorgonia`

Closes #277

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,6 @@ require (
 	gopkg.in/cheggaaa/pb.v1 v1.0.27
 	gorgonia.org/cu v0.9.0-beta
 	gorgonia.org/dawson v1.1.0
-	gorgonia.org/gorgonia v0.9.1
 	gorgonia.org/tensor v0.9.0-beta
 	gorgonia.org/vecf32 v0.7.0
 	gorgonia.org/vecf64 v0.7.0


chewxy:
@durp how confident are you of this solution? (sorry, not being difficult, but I personally don't quite fully understand gomod yet)

durp:
As you may have noticed, the module designation of `gorgonia` vs `gorgonia.org/gorgonia` does not seem to affect the ability to compile `gorgonia` itself, but the former has a subtle effect `gorgonia.org` (v9.1, I might add!), appears as a dependency in the `requirements` section of `go.mod`

That is because all the imports in say, `examples` are using the fully qualified name... which `go mod` assumes (since they ask for `gorgonia.org` and not `gorgonia`) is from outside the module.

The source matches well enough (though it is not `master`!! (potential surprise)), so everything compiles even with modules enabled.

The problem occurs when another project tries to `go get` as mention in #227 - now `go mod` is parsing the dependencies and happens upon `gorgonia` and gives the ubiquitous and relatively uninformative "unexpected module path" error.

I am not familiar enough with the `go mod` underpinnings yet to tell if this is *expressly* because `gorgonia` is mentioned in the requirements or the solver just says, "this doesn't add up". In any case, it quits real quick.

I tried various things (looking for code that had "bad" import paths within `gorgonia`, etc.) before coming to to the `doh` realization the module name itself was fouling the pipes.

To test the fix, I verified that I could still compile and run all tests in `gorgonia` and that I could `go get` and import `gorgonia` into another project (a fork of `chewy/lingo`, FWIW) without issue.

TLDR; I am confident of this fix

chewxy:
Thank you

PR #294, opened 2019-06-07
Maxpool ceilmode

owulveryck:
**TL;DR:** This PR is a draft to open the discussion about a possibility to break the API for the Maxpool operator to add a "ceil mode".

**Longer:** 

From version 1.5, onnx supports a "ceil mode" attribute that has an impact on the calculus of the output shape on the Maxpool operator. (see [this disccusion](https://github.com/onnx/onnx/issues/549) for more info).

By now, I did not find any model that is using this mode, nevertheless, I've implemented a basic implementation to do different tests (one again **I am only using this branch for testing purpose**.)

The problem is that it breaks the API by adding a new attribute to the `MaxPool2D` function; another option would be to keep this function as-is and silently set the `ceilMode` to false and to create a new transition function `MaxPool2DCeil` but it do not really like this idea.

WDYT?

### DIFF
--- a/known_issues_test.go
+++ b/known_issues_test.go
@@ -129,7 +129,7 @@ func TestIssue273_maxpool_pads(t *testing.T) {
 			48, 49, 49, 49, 45, 46, 47, 48, 49, 49, 49, 45, 46, 47, 48, 49, 49, 49,
 		}))
 
-	y, err := MaxPool2D(x, []int{3, 3}, []int{2, 2}, []int{1, 1})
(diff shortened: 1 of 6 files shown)


owulveryck:
One more thing: this branch also includes in shadow the premises of implementation of `dilation`

chewxy:
I like it. Perhaps we should think of using `...interface{}` for `MaxPool2D` for now. This allows for a future transition to using functional options like 

```
type NNOpt func(Op) error
```

which would fit into the building of golgi nicely in the future

zimenglan-sysu-512:
any progress updates?

PR #295, opened 2019-06-09
feat(maxpool): explicit and asymmetric padding

owulveryck:
This PR is an attempt to deal with explicit and asymmetric padding for the MaxPool operator.

This does not break the current API and has been tested against the ONNX tests.
I think this has no impact on the gradient computation, but I am not sure.


chewxy:
LGTM. I think names of the pads are a bit confusing. Cardinal direction (North, south east west) might be easier? Thoughts?

PR #298, opened 2019-06-25
chore: doc about cuda

owulveryck:


owulveryck:
If my understanding is correct, the `UseCudaFor` method has disappeared since v0.9. 
The idiomatic way to use CUDA is now to use the `nnops` package. This documents it.

On top of that, I have tweaked the README.md to use the [`embedmd`](https://github.com/campoy/embedmd) tool to ease the maintenance of the by reusing the real code.

PR #300, opened 2019-06-29
Added BroadcastAdd and BroadcastMul as operations

chewxy:


chewxy:
I found a few uncaught errors in the Broadcast operation. Thoughts on the API?

chewxy:
p/s: I'm aware we've talked about adding this, and the `Broadcast()` function is an amazing API

And this seems to be a bit backtracking on my previous thoughts in #223 but @owulveryck 's new `Broadcast` API makes perfect sense still in terms of extensibility. This `BroadcastAdd` and `BroadcastMul` are added for convenience - instead of having to always redefinte `Add`

owulveryck:
I agree that the two convenient function will help to have a readable code.
 It’s better to have that mainstream instead of making the client do that in their package.

PR #302, opened 2019-07-01
Gap operator

owulveryck:
This is an initial implementation of the Global Average Pool Operator.

It probably requires some optimization.
On top of that this operator does not handle gradient computation yet. This is why this PR is in a draft.


owulveryck:
@chewxy are you ok to merge this one without the implementation of the `ADOp` interface?

chewxy:
Yep

PR #307, opened 2019-07-23
V0.10.0 working

chewxy:
This will be a huge change. And this PR is going to be the living document.


chewxy:
Ugh and this is why you don't refactor at 3am. `values` `types` and `shapes` should be moved to `internal`

PR #308, opened 2019-07-31
feat: dot representation of a graph as an external package

owulveryck:
This PR holds some code I made a while ago.
The ultimate goal is to remove the `ToDot` from the main package.
Using an external package gives the ability to easily adapt the dot output without breaking any code mainstream.

I may add a notion of `groups` within the node; This is useful for writing the dot output, but may also be useful for encoding into `onnx` (an operator can be composed of a group of node, and we need to identify it to encode properly)

### DIFF
--- a/encoding/constants.go
+++ b/encoding/constants.go
@@ -1,0 +1,19 @@
+package encoding
+
+// Cluster represent a group of nodes that are similars
+// It is used for the grapviz generation
+
+const (
+	// UndefinedCluster ...
+	UndefinedCluster GroupID = iota
+	// ExprGraphCluster is the default cluster
+	ExprGraphCluster
+	// ConstantCluster is the group of nodes that represents constants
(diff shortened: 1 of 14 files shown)


chewxy:
LGTM. Here's a plan:

1. Merge this to master but not do a point release for now
2. Pull from master to v0.10.0-working

Are you fine with that?

owulveryck:
I am ok with this for this specific commit.
But this can be annoying if I start to clean the main tree by removing the todot methods.
Let’s do that for this specific commit that will not break anything and see later how to deal with the rest :)

chewxy:
Cool. The main reason I am asking this is because we want to start encouraging people to use `encoding/dot`

owulveryck:
I am closing this PR. A new one will be created to merge the encoding/dot to master

PR #309, opened 2019-08-03
init of the encoding/dot package

owulveryck:
This PR does not introduce any regression.

It just adds some deprecated notes in some of the export function to warn the end-user that the ToDot exports may disappear.

A kind of abstraction has been initiated in the encoding package. This package is `internal` because I don't see any good reason to make it available outside of Gorgonia. On top of that, this will allow the API to change when we'll start the ONNX export.


owulveryck:
This is an example generated from the linear-regression example of the core package.
linear-regression

owulveryck:
The last commit improves the graph and does a subgraph when a node is part of the `ExprCluster` **and** another group.

I did a sample with the convolution operator:

bla

PR #314, opened 2019-09-02
small fixes pointed by golint

owulveryck:
The purpose of this PR is to do small fixes pointed by the linter.
Most of them are about documentation.

see https://go-lint.appspot.com/gorgonia.org/gorgonia

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -1,15 +1,18 @@
 package gorgonia
 
+// Batched interface describes ny object that can process batch work
 type Batched interface {
 	WorkAvailable() <-chan struct{}
 	DoWork()
 }
 
+// BatchedBLAS interface describes ny object that can process BLAS work in batch
 type BatchedBLAS interface {
 	Batched
 	BLAS
 }
 
+// BatchedDevice ...
 type BatchedDevice interface {
 	Batched
 	Retval() interface{}
--- a/batch.go
+++ b/batch.go
@@ -12,7 +12,7 @@ type BatchedBLAS interface {
 	BLAS
 }
 
-// BatchedDevice ...
+// BatchedDevice is the superset of BatchedBLAS and the batched CUDA workflow.
 type BatchedDevice interface {
 	Batched
 	Retval() interface{}
--- a/batch.go
+++ b/batch.go
@@ -1,12 +1,12 @@
 package gorgonia
 
-// Batched interface describes ny object that can process batch work
+// Batched interface describes any object that can process batch work
 type Batched interface {
 	WorkAvailable() <-chan struct{}
 	DoWork()
 }
 
-// BatchedBLAS interface describes ny object that can process BLAS work in batch
+// BatchedBLAS interface describes any object that can process BLAS work in batch
 type BatchedBLAS interface {
 	Batched
 	BLAS
--- a/complex_test.go
+++ b/complex_test.go
@@ -25,16 +25,16 @@ func TestWeirdNetwork(t *testing.T) {
 	var p *Node
 	var q *Node
 
-	e_x := NewMatrix(g, Float64, WithShape(xs, embeddingDims), WithName("x embeddings"), WithInit(GlorotU(1)))
-	e_p := NewMatrix(g, Float64, WithShape(ps, embeddingDims), WithName("p embeddings"), WithInit(GlorotU(1)))
-	e_q := NewMatrix(g, Float64, WithShape(qs, embeddingDims), WithName("q embeddings"), WithInit(GlorotU(1)))
-	w0_x := NewMatrix(g, Float64, WithShape(hiddenSize, xFeats*embeddingDims), WithName("layer0 weights for x"), WithInit(GlorotU(1)))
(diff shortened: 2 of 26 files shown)


owulveryck:
Once the work is done, the only messages that should remain are those related to Public functions returning private objects.
But this goes beyond linting and the fixes should be postponed to a release (it will probably have an impact on the API)

utils.go
```diff
@@ -139,10 +139,8 @@ func hasInf(v Value, dev Device) bool {
 	switch vt := v.(type) {
 	case *F64:
 		return false
```
chewxy: This is incorrect

utils.go
```diff
@@ -139,10 +139,8 @@ func hasInf(v Value, dev Device) bool {
...
-		return math.IsInf(float64(*vt), 0)
```
chewxy: This is correct. 

I think I added the `return false` during some debug

batch.go
```diff
@@ -1,15 +1,18 @@
...
 	DoWork()
 }
 
+// BatchedBLAS interface describes any object that can process BLAS work in batch
 type BatchedBLAS interface {
 	Batched
 	BLAS
 }
 
+// BatchedDevice ...
```
owulveryck: I don't know what this interface represents

complex_test.go
```diff
@@ -25,16 +25,16 @@ func TestWeirdNetwork(t *testing.T) {
 	var p *Node
 	var q *Node
 
-	e_x := NewMatrix(g, Float64, WithShape(xs, embeddingDims), WithName("x embeddings"), WithInit(GlorotU(1)))
-	e_p := NewMatrix(g, Float64, WithShape(ps, embeddingDims), WithName("p embeddings"), WithInit(GlorotU(1)))
-	e_q := NewMatrix(g, Float64, WithShape(qs, embeddingDims), WithName("q embeddings"), WithInit(GlorotU(1)))
-	w0_x := NewMatrix(g, Float64, WithShape(hiddenSize, xFeats*embeddingDims), WithName("layer0 weights for x"), WithInit(GlorotU(1)))
-	w0_p := NewMatrix(g, Float64, WithShape(hiddenSize, pFeats*embeddingDims), WithName("layer0 weights for p"), WithInit(GlorotU(1)))
-	w0_q := NewMatrix(g, Float64, WithShape(hiddenSize, qFeats*embeddingDims), WithName("layer0 weights for q"), WithInit(GlorotU(1)))
+	eX := NewMatrix(g, Float64, WithShape(xs, embeddingDims), WithName("x embeddings"), WithInit(GlorotU(1)))
```
owulveryck: I am aware that the underscore may have a meaning, but the linter doesn't agree

debug.go
```diff
@@ -32,12 +33,9 @@ var (
 	allocatorDev      = false
 )
 
-var READMEMSTATS = true
-
+// TABCOUNT is a global flag used when debugging
```
owulveryck: Don't know what it does

debug.go
```diff
@@ -32,12 +33,9 @@ var (
 	allocatorDev      = false
 )
 
-var READMEMSTATS = true
```
owulveryck: Did not find any reference anywhere, so I consider it as dead code

debug.go
```diff
@@ -32,12 +33,9 @@ var (
...
-
+// TABCOUNT is a global flag used when debugging
 var TABCOUNT uint32
 
-var TRACK = false
```
owulveryck: Did not find any reference anywhere, so I consider it as dead code

engine.go
```diff
@@ -5,13 +5,12 @@ import (
 	"gorgonia.org/tensor"
 )
 
-type SymbolicEngine struct{}
```
owulveryck: Did not find any reference anywhere, so I consider it as dead code.

noextern.go
```diff
@@ -72,6 +72,7 @@ func (m *ExternMetadata) Transfer(toDev, fromDev Device, v Value, synchronous bo
 	return v, nil
 }
 
+// Reset is a noop function for compatibility with the interface
```
owulveryck: which interface btw?

op.go
```diff
@@ -96,12 +96,6 @@ type BinaryOp interface {
 	IsBinary() bool
 }
 
-type BestDoer interface {
```
owulveryck: Already discussed: Did not find any reference anywhere, so I consider it as dead code

batch.go
chewxy: BatchedDevices are the superset of BatchedBlas and the batched CUDA workflow.

debug.go
chewxy: readmemstats is an old variable used to dump memory statistics. Not used any more

engine.go
chewxy: `SymbolicEngine` is for the upcoming v0.10.0 API

noextern.go
chewxy: It's a fake "interface". Both `// +build cuda` and `// +build !cuda` requires a `Reset()` function,

chewxy:
can I merge this?

owulveryck:
Good to merge. Thank you.

Issue #311, opened 2019-08-17
Fix build at Raspberry Pi


diegobernardes:
I had some problems to use `gorgonia` with a Raspberry Pi. This are the steps that I did to solve the problem.

1. Update the dependencies to use `github.com/chewxy/math32 1.0.1`.
2. Update the `mathutils` files. The current `asm` implementation is not compatible with the ARM, so I had to skip it with `noasm`, but the problem is that it also skips the pure Go version as well:

```
go build -tags='noasm' main.go
# gorgonia.org/gorgonia
../../../gorgonia/bitmap.go:17:10: undefined: divmod
../../../gorgonia/bitmap.go:35:16: undefined: divmod
../../../gorgonia/bitmap.go:45:16: undefined: divmod
../../../gorgonia/bitmap.go:55:16: undefined: divmod
../../../gorgonia/operations.go:549:20: undefined: divmod
```

And if I add the tag `wasm`, that is used to the pure Go version, other errors happen because it's used internally at the Go runtime to build `wasm`.

cpllbstr:
Faced that error today with Jetson Nano. The solution is pretty simple. There is a error in **mathutils.go** there is function **divmod** that is declared but not implemented. You can simply delete the comment line `// +build wasm` in mathutils_go.go and **divmod** will work fine.

bezineb5:
#321 should have fixed this issue.

PR #315, opened 2019-09-03
Fix when noasm tag is used

bezineb5:
Compilation fails when noasm is used, as this method is not defined in that case.
This is a fix for #311


chewxy:
LGTM. I will add the `noasm` to the yaml as well

bezineb5:
Thanks!

chewxy:
See #316 . I wasn't able to add directly to this PR... but your contributions remain.

PR #318, opened 2019-09-04
Go report

owulveryck:


op_nn.go
```diff
@@ -348,7 +346,7 @@ func (op im2colOp) do(prealloc, input Value) (retVal Value, err error) {
 			imStart := i * batchStrideIm
 			colStart := i * batchStrideCol
 
-			if imEnd = imStart + batchStrideIm; imEnd >= len(imData) {
```
owulveryck: This is wrong, I will fix that.

equalities.go
```diff
@@ -98,7 +98,7 @@ func tensorClose(a, b tensor.Tensor) bool {
 		if len(aFs) != len(bFs) {
 			return false
 		}
-		aFs = aFs[:len(aFs)]
+		aFs = aFs[:]
```
chewxy: actually this is required for the bounds check removal

`aFs` and `bFs` are bound to `len(aFs)`, so the compiler knows that these are the same length and thus do not issue bounds check instructions. This speeds things up

owulveryck: TBH, I did not think about it (I even forgot about boundary checks, since I didn't deal with it since the.... well a long time) ; I simply applied the hint `gofmt -s` gave me.
Your comment is actually very relevant, and I wanted to see if the hint was correct.

It looks like the go compiler is smart enough to not put boundary check instruction when ` aFs[:]` is called.

this [article](https://medium.com/@brianblakewong/optimizing-go-bounds-check-elimination-f4be681ba030) gives a hint to display the boundary checks by adding a `gcflags` to the go build.

I did it and it does not show any boundary check on equalities with the new code: 
```
✗ grep "aFs =" equalities.go
                aFs = aFs[:]
                aFs = aFs[:]
✗  go build -gcflags="-d=ssa/check_bce/debug=1" 2>&1 | grep -q equalities || echo ok
ok
```

One or the other syntax sounds perfectly equals.

chewxy:
lmk if this is good to merge

owulveryck:
Good for me

PR #319, opened 2019-09-04
fix: receiver name has changed, change its reference

owulveryck:


owulveryck:
I introduced a bug in the refactoring; this has not been seen by my linter nor by the CI because I did not try to build with CUDA until now.
With this PR, the build is OK.

PR #322, opened 2019-09-07
Fix stabilization for Log(x+k), k!=1

bezineb5:
This is a fix for the following issue:
When there is an operation like Log(x+2), the stabilization will replace it by Log(nil)
See unit test for an example.
Thanks!


chewxy:
What an excellent bugfinding and fixing!

PR #331, opened 2019-10-08
feat: new basic test for the Use function

owulveryck:


owulveryck:
This is a basic test to validate that the `Use` function is actually doing something.
Side effect, it increases the code coverage :D

PR #337, opened 2019-10-13
fix(gomachine): do not overwrite the channel if it already exists

owulveryck:


.travis.yml
```diff
@@ -5,9 +5,8 @@ branches:
   only:
     - master
 go:
-  - 1.10.x
-  - 1.11.x
   - 1.12.x
+  - 1.13.x
   - tip
```
owulveryck: @chewxy there is an incompatibility with go < 1.12.
I did not really investigate the root cause, but It looks like the latest gonum it using [reflect.MapIter](https://golang.org/pkg/reflect/#MapIter) that is only accessible in 1.12+. (see [build](https://travis-ci.org/gorgonia/gorgonia/jobs/597200820) for more details)

owulveryck: As seen on slack, let's go for go >= 1.12

PR #339, opened 2019-10-14
Asm arm build tags

owulveryck:
If I am not wrong, the ASM version works only on amd64.
Therefore using an extension and a negative build tag makes this work on all cases.


chewxy:
@jokebroker if all good pls merge next time

Issue #323, opened 2019-09-12
SymDiff failure for StableSoftMax


bdleitner:
Trying to use SoftMax-BinaryXent as a cost function, I run into errors when computing the Gradient:

I can reproduce with
```
package main

import (
  "gorgonia.org/gorgonia"
  "gorgonia.org/tensor"
  "log"
)

func main() {
  g := gorgonia.NewGraph()
  var params gorgonia.Nodes
  x := gorgonia.NewTensor(g, tensor.Float64, 2, gorgonia.WithShape(1, 10), gorgonia.WithName("x"))
  w := gorgonia.NewTensor(g, tensor.Float64, 2, gorgonia.WithShape(10, 7), gorgonia.WithInit(gorgonia.GlorotN(1.0)))
  params = append(params, w)

  p := gorgonia.Must(gorgonia.Mul(x, w))
  sf :=gorgonia.Must(gorgonia.StableSoftMax(p))
  y := gorgonia.NewTensor(g, tensor.Float64, 2, gorgonia.WithShape(1, 7), gorgonia.WithName("y"))

  xe := gorgonia.Must(gorgonia.BinaryXent(sf, y))
  cost := gorgonia.Must(gorgonia.Mean(gorgonia.Must(gorgonia.Sum(xe, 1))))
  if _, err := gorgonia.Grad(cost, params...); err != nil {
    // Errors out here
    log.Fatalf("Could not compute Gradient: %v", err)
  }

  vm := gorgonia.NewTapeMachine(g, gorgonia.BindDualValues(w))
  solver := gorgonia.NewRMSPropSolver(gorgonia.WithBatchSize(float64(1)))

  gorgonia.Let(x, tensor.NewDense(tensor.Float64, tensor.Shape{1, 10}, tensor.WithBacking([]float64{1, 1, 1, 1, 1, 1, 1, 1, 1, 1})))
  gorgonia.Let(y, tensor.NewDense(tensor.Float64, tensor.Shape{1, 7}, tensor.WithBacking([]float64{0, 1, 0, 0, 0, 0, 0})))

  if err := vm.RunAll(); err != nil {
    log.Fatalf("Runall failed: %v", err)
  }

  solver.Step(gorgonia.NodesToValueGrads(params))
  vm.Reset()
}
```

This gets the error:
```
[...]

PR #343, opened 2019-10-20
Correct shape/type inference for Max/Sum

### LINKED ISSUE #326: Max Panics for dimensions > 2 (and along != everything)
I first found this issue when playing with `LogSumExp` (which seems to have other issues, but I'll file something separate if I can confirm there's more of a problem than what the Max issue exposes).

If the dimensions of the input are > 2 and the `along` parameter is neither empty not all dimensions, Max panics.  For repro, see the code below.

The issue appears to be that the `maxOp`'s `Type()` method reduces the number of dimensions by 1 (which already seems wrong as `along` could have more than 1 entry and still not be everything).  On the other hand, `InferShape` simply replaces those axes in the `along` parameter with 1's.

When the resulting node is constructed, there is a `WithShape` option applied (op.go:202) with the shape from `InferShape`, but the node type has already been set and though there is no shape, the tensor number of dimensions is now `n-1`.  Thus, the `WithShape` function panics as `acceptVec` and `acceptScalar` will both be false, but the number of dimensions do not match.

A few potential solutions come to mind (and there may be others I haven't thought of.  I considered seeing about becoming a contributor, and still might, but I still don't know enough about the deep internals to know which approach to take here):

1. Have `Max` (and other reduction ops like `Sum` which also has a similar issue) preserve the number of dimensions.  This would mean not having the `Type()` method reduce the number of dims by 1.
1. Have `InferShape` for the reduction operations remove the axes in the `along` list rather than convert them to 1s.
1. Have `WithShape` be tolerant of mismatched dimensions if the extra dimensions have length 1.

It seems to me that #1 is the best option.  For reduction operations, the values are still there even if they've been reduced along a dimension, and so for any future operations (which may involve broadcasting), it makes sense to keep the number of dimensions the same.

If you agree with that approach, I'm willing to take on making the changes (full disclosure: it would be my first ever contribution to an open source project).

```
package main

import (
	"fmt"

	"gorgonia.org/gorgonia"
	"gorgonia.org/tensor"
)

func main() {
	// This works fine
	doMax(tensor.Shape{5, 7})
	// This panics with:
	// Matrix float64, has 2 dimensions(Shape: ()). Input shape is (1, 7, 11), which has 3 dimensions
	doMax(tensor.Shape{5, 7, 11}) // panics

	// This also panics (if we comment out the above so we can get here)
	// Tensor-3 float64, has 3 dimensions(Shape: ()). Input shape is (1, 7, 11, 13), which has 4 dimensions
	doMax(tensor.Shape{5, 7, 11, 13}) // panics
}

func doMax(shape tensor.Shape) {
	g := gorgonia.NewGraph()
	in := gorgonia.NewTensor(
		g,
		tensor.Float64,
		len(shape),
		gorgonia.WithShape(shape...),
		gorgonia.WithInit(gorgonia.RangedFrom(0)))
	fmt.Println(in.Shape())
	for i := 0; i < len(in.Shape()); i++ {
		out, err := gorgonia.Max(in, i)
		if err != nil {
			fmt.Println(err)
			continue
		}
		fmt.Printf("Max(%d) has shape: %v\n", i, out.Shape())
	}
}
```

bdleitner:
Refactor the max/sum ops to share common code.  Have the type/inferShape/Do methods behave in a consistent manner:

* Dimensions specified in the `along` parameter are reduced to size 1, but not necessarily removed.
* If all dimensions are included, the result will be a scalar.
* If all dimensions but 1 are included, the result is a vector, regardless of which dimension is left intact.

Tests verify that the resulting nodes have the expected shape.

Note: While here, fix a warning on Max's SymDiff where retVal[0] is set when retVal has not been initialized.  I wonder if this is related to #323 where SymDiff for StableSoftMax (which uses Max) was failing with a panic (probably not, as the error message there seems unrelated, but probably a good fix anyway).

Closes #326


bdleitner:
Notes:

* There is one test failing: TestRepeatOpDoDiff, but this seems unrelated as I see the same failure in a clean branch.
* I had to modify go.mod to use gorgonia.org/tensor v0.9.2 (to pick up the fix for [issue 47](https://github.com/gorgonia/tensor/issues/47), but my Goland IDE insists on making other changes... I haven't figured out (yet) how to turn off the auto-generation.

chewxy:
Looking into that failing test now

bdleitner:
After further review, it looks like my "clean" client was not actually clean, and looking into the bug it seems like something that is related to my change.

Since my local branches are not in the state I thought they were and there have been other submits in the interim, rather than try and clean this up, I'm just going to start over with a fresh fork.  Expect a new pull request coming soon (depends on how complex fixing the other test is and how much time I'll have during the week).

PR #346, opened 2019-10-28
Ensure consistent shape behavior for type/inferShape on reduction ops

bdleitner:
Refactor the max/sum ops to share common code. Have the type/inferShape/Do methods behave in a consistent manner:

* Dimensions specified in the "along" parameter are reduced to size 1, but not removed. (Note: this caused TestRepeatOpDoDiff, but this version fixes it.  Perhaps we should make preserving the size-1 dimensions an option of the reduction op?)
* If all dimensions are included, the result will be a scalar.
* If all dimensions but 1 are included, the result is a vector, regardless of which dimension is left intact.

Tests verify that the resulting nodes have the expected shape.

Note: While here, fix a warning on Max's SymDiff where retVal[0] is set when retVal has not been initialized.  I wonder if this is related to gorgonia#323 where SymDiff for StableSoftMax (which uses Max) was failing with a panic (probably not, as the error message there seems unrelated, but probably a good fix anyway).

Closes gorgonia#326


bdleitner:
Ok, fixing that other test was easier than I thought it might turn out to be.
There is a somewhat open question about how to handle the leftover size-1 dimensions.  This code preserves them, unless the resulting shape has 0 or 1 non-size-one dimensions, in which case it becomes a scalar or vector, respectively.  For other outputs, the size-1 dimenions stay.

An alternative approach would be to have them always removed.  I think both are reasonable, but I'm leaning toward saying that leaving them in is preferable as operations like Max or Sum are useful building blocks for combining with things like Add/Sub, as happens with e.g. StableSoftMax.  Having the dimension present would make broadcasting simpler.

I'm happy to rework this to have the size-1 dimensions consistently eliminated instead, I'll defer to your judgement on that.

chewxy:
Sorry I was running GopherConAU. WIll look into this on the weekend

bdleitner:
I'm not sure what to do about the go.mod and go.sum files.  GoLand seems to insist on rebuilding those, and I haven't yet figured out how to avoid it.   I'm happy to revert the changes and see if I can find how to make GoLand do what I want (or find another workaround), but I'm not sure what the correct course of action is at this point.  Please advise.

chewxy:
hmmm I'm not sure if I fixed it.. gimme afew mins

chewxy:
Thank you @bdleitner

PR #350, opened 2019-11-08
Fix go.mod

mattn:
Fix dependency for gota.

https://github.com/go-gota/gota/issues/100


chewxy:
@jokebroker check and merge when ready pls thx

PR #359, opened 2020-01-04
Fixed a bug in Reshape, where a slice of a slice cannot be reshaped.

chewxy:
This is fixed in two ways:
    1. If the input is a `View` then it is materialized. This
    requires more memory, but it is more worth it to allocate than to
    stress over calculating  how much extra overhead to allocate for
    sharing memories
    2. `ShallowClone` is fixed in package `tensor` to make sure that
    views are correctly shallow cloned as well


chewxy:
https://github.com/gorgonia/tensor/pull/56

Additional notes: this fix is a hack. A better fix is to not have to call `Materialize`, but instead allow for reshaping non-contiguous slices in `tensor`.

PR #361, opened 2020-01-09
Refactor monads

chewxy:


ermagerdmonards.go
```diff
@@ -28,27 +27,32 @@ type Errer interface {
 	Err() error
 }
 
+// Mker is an interface of any Input that can make a new version of itself
+type Mker interface {
```
owulveryck: I wonder what is the use-case for this interface :D

chewxy:
```
type Carrier struct {
    *Node
    message string
}

func (c Carrier) Mk(ins ...Input) Input {
    return Carrier{ins[0].(*Node), c.message}
}

func example() {
    x := NewMatrix(...)
    y := NewMatrix(...)
    add := Lift2(Add)
    z := add(Carrier{x, "hello"}, y)
    msg := z.(Carrier).message
}
```

owulveryck: TBH, I still don't get the point for the `Mk` with this example.
I see it used within the `TransformResult`, but, this can be avoided by using the function directly?

Issue #301, opened 2019-07-01
Add More Tests For Broadcast Operations


chewxy:
In [operations_broadcast_test.go](https://github.com/gorgonia/gorgonia/blob/master/operations_broadcast_test.go) there are some tests that are not yet filled in. The point is to test that broadcasting works for different shapes. The semantics of broadcast probably isn't clear, so please do send me a message for anything. 

This is a good first issue for anyone looking to get interested

blackrez:
I'm volunteer for this issue, my skills in ML is very low and I want to improve them.

chewxy:
FYI the API for broadcast operations are now autogenerated in the next branch.

evanmcclure:
Is this issue still available?

chewxy:
Yes!

chewxy:
In the new branch (golgi-supports) you will see all the autogenerated APIs. It'd be good to test those

evanmcclure:
Thanks. I'll take it if @blackrez doesn't want it.

blackrez:
If @evanmcclure want to work on it, it's ok for me.
Actually, I'm busy with my new project based on gorgonia. :simple_smile:

supun:
@evanmcclure Is this still open ?

chewxy:
@supun The answer is yes. The tests added in #367 solved a bunch of issues, but also raised a bunch of others. If you want, you can tackle the rest of the operations. The tests will most likely fail, but I will still merge them if they fail in the correct way.

I am currently working on the concrete semantics of broadcast. Once that's done, all the failing tests should pass

chewxy:
Go ahead

*Xuanyi Chew*
*Follow: @chewxy <https://twitter.com/chewxy> on Twitter*
*Watch: my talk on Closure Calculus -  An Alien Lambda Calculus
<https://www.youtube.com/watch?v=ogXlQf8lDD4>*
*Read:* *my blog <http://blog.chewxy.com> *
*+61403928398*

kai-xlr:
Okay cool 😎

thani-ath-nain:
Seems like an interesting issue, and I don’t see any currently open PRs linked to it. I’ve worked with Go and ML for about 3 years, and I’d like to pick this up. I’ll open a PR today.

PR #367, opened 2020-01-26
Test broadcast add

jorgecarleitao:
See #301 

Note that this branch currently fails, but I believe that it is due an implementation error in the broadcasting. Specifically, I was unable to make the following operation:

given a tensor `a` with shape `(2,)` and a tensor `b` with shape `(2,2,2)`, broadcast-add them, such that the result `c` has shape `(2,2,2)`. This stems from the fact that when I try to broadcast `a` into shape `(2,2,2)` (using `left=[]byte{1, 2}`) in the broadcastAdd, it panics.

This situation is demonstrated on the second commit of this MR.

This behavior is not consistent with when we do the same operation where `b` has shape `(2,2)` and `a` is broadcasted using `left=[]byte{1}`, which is valid (as per test named `"vec-mat"`).

Note that I am assuming that we are following the same rules as [the broadcasting rules of numpy](https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html)


chewxy:
lgtm. its 11pm here ... will merge tomorrow. did this include a fix?

jorgecarleitao:
> lgtm. its 11pm here ... will merge tomorrow. did this include a fix?

it does not include a fix. I wanted to check with someone more knowledgeable first that there is indeed an implementation error before proceeding with a fix.

jorgecarleitao:
The issue seems to originate from https://github.com/gorgonia/tensor/blob/master/shape.go#L265, which handles row-vectors, column-vectors and the axis=1 in a special way that, when called by https://github.com/gorgonia/gorgonia/blob/master/op_tensor.go#L267, leads to the error on logged in travis.

chewxy:
You have that correct! The definition and operations on shapes have not been in sync for this issue. I don't quite have the time this week (working on some Golgi API and #362 ). Do you want to handle this?

chewxy:
@jorgecarleitao ^

jorgecarleitao:
I do not think that I can commit to the fix right now, as this is a task a bit bigger than what I antecipated. On the weekend I may have some time.

chewxy:
weekend works :)

chewxy:
OK I have looked at this problem. More notes to come, and opinions solicited

chewxy:
@jorgecarleitao has highlighted a major issue in Gorgonia's approach to things. This document will attempt to synthesise the various viewpoints and the tensions that exist.

# Gorgonia's Position #

The position of Gorgonia (the library) is that Gorgonia is a library that enables **clear** APIs for creating and manipulating mathematical expressions for the purposes of deep learning.

The mathematical objects we wish to manipulate are the following:

* mathematical expression terms
    - nodes (tensors)
	- formation rules of the tensors (Op)

Clarity means no magic. Magic can largely be described as "having unintuitive semantics". The goal is of course, for a person who has mathematical knowledge to be able to trivially pick up and use Gorgonia. There is also the semantics of Go to contend with - hence the semantics of the library are largely aligned to the usual patterns of Go.

An example is where functions return `(*Node, error)`. This is idiomatic Go - ["don't panic"](https://github.com/golang/go/wiki/CodeReviewComments#dont-panic) is the proverb regarding error handling.

# The Problem #

The problem is enumerated well in this PR. However, the problem merely scratches the surface of the true issue. In a subsection, I provide a technical solution. However, **I think a technical solution is insufficient, and we will eventually face problems around broadcasting again**.

The basic premise of the problem is best summarized in the following concrete examples:

* A vector (e.g. with shape (2,)) is unable to be broadcasted to a 3-tensor of a compatible shape (e.g. with shape (2, 3, 4)). :x:
* A vector (e.g. with shape (2,)) is able to be broadcasted to a matrix of a compatible shape (e.g. with shape (2,3)) - the example can be seen [here](https://pkg.go.dev/gorgonia.org/gorgonia?tab=doc#BroadcastAdd). :white_check_mark:

Numpy is able to do so - following the [broadcasting rules](https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html). Therein lies the tension. The Numpy API has normalized broadcasting as an implicit operation since 1999.

Now, to discuss the finer points, let's set some definitions:

* **inner(most) dimension(s)** - given a shape, the inner dimensions tend to the right. e.g. in a shape (2,3,4), the inner dimensions are (3, 4). The innermost dimension is 4.
* **outer(most) dimension(s)** - given a shape, the outer dimensions tend to the left . e.g. in a shape (2,3,4), the outer dimensions are (2, 3). The outermost dimension is 2.
* **vector** - a `tensor` of rank-1. e.g. (2), (3) ... Vectors will be written `[...]`
* **column vector** (or **colvec**) - a `tensor` of rank-2 (i.e. a matrix) with the inner dimension as 1. e.g. (2, 1), (3, 1)...
* **row vector** - a `tensor` of rank-2 (i.e. a matrix) with the outer dimension as 1. e.g. (1, 2), (1, 3)...
* **matrix** a `tensor` of rank-2, with arbitrary inner and outer dimensions. e.g (1, 2), (2, 1), (2, 3)... Matrices will be written in full notation in this doc. This includes colvecs and rowvecs.

## The Technical Solution ##

<details>
<summary>Click to expand</summary>

Having discovered this issue, the technical solution is quite trivial - simply reshape a vector (and only a vector) to fit the broadcast axes.

### Example ###

Given
```
// a is a (2, 3, 4) 3-tensor
a =
	⎡1  2  3  4 ⎤
	⎢0  0  0  0 ⎥
	⎣5  6  7  8 ⎦
	⎡10 20 30 40⎤
	⎢0  0  0  0 ⎥
	⎣50 60 70 80⎦

// b is a (4) vector
b = [100 200 300 400]
```

In this example, `BroadcastXXX(a, b, nil, []byte{0,1})` would yield an intermediate `b_`  which is this:


```
b_ =
	⎡100 200 300 400⎤
	⎢100 200 300 400⎥
	⎣100 200 300 400⎦
	⎡100 200 300 400⎤
	⎢100 200 300 400⎥
	⎣100 200 300 400⎦
```

The way this is done is that `b` is first reshaped into (1, 1, 4). This can be done with something like the following code:

```
func magic(a, b *Node, along [2][]int) (*Node, *Node, error) {
	d := a.Shape().Dims()
	sort.Sort(along)
	// check
	var unbc, prev int
	for _, v := range along{
		if v >= d {
			// error
		}
		if v - prev > 2{
			unbc++
		}
		prev = v
	}
	if unbc > b.Dims() && !b.IsVector() {
		// error: number of unbroadcasted dims is bigger than the dimensions of b
	}

	newshape := make(tensor.Shape, d)
	for i := range newshape{
		for _, j := range along {
			if j == i {
				newshape[i] = 1
			}
		}
	}

	var lastused int
	for i := range newshape {
		if newshape[i] == 0 {
			newshape[i] = b.Shape()[lastused]
			lastused++
		}
	}

	// reshape to new shape
	b_, err := Reshape(b, newshape)


	// ... other code
	return ___, b_
}

func Broadcast(a, b *Node, pat BroadcastPattern){
	magic(a, b, pat.on())

	// remaining Broadcast code
}

```

</details>


The technical solution is trivial. It's also trivial to write tests that ensure it's correct. However, the bigger problem is whether this fits into the Gorgonia vision.

For now, let us go back to the original reasons to have the `BroadcastXXX` functions, then let's look at the various early variations, as it may be worth revisiting the various implementations.


# Current State of The World #

Numpy has introduced new semantics to mathematical objects by introducing broadcasting.

<details>
<summary>Click to expand </summary>

Recall that the main mathematical objects we're interested in manipulating are tensors. In the linear algebra world, there is no notion of broadcasting. The following things cannot be computed:

```
a = ⎡1  2  3  4 ⎤      b = [10 20 30 40]
    ⎢0  0  0  0 ⎥      c = [100 200 300]
    ⎣5  6  7  8 ⎦
```

In linear algebra, `a + b` is not well defined. Neither is `a + c` and `b + c`. However, in Numpy, `a + b` does yield a result, while `a + c` and `b + c` causes an error. This is because Numpy allows for broadcasting, and that requires a new set of assumptions.

The reason why `a + b` is ill-defined in the linear algebra world is because `a`, `b` and `c` are different types. For this section, let's give the types names (using Go type assertion notation):

* `a.(matrix3_4)`
* `b.(vector4)`
* `c.(vector3)`

The reason why it works in Numpy is due to a recasting of semantics. Instead of understanding `a` as `matrix3_4`, Numpy allows for it to be thought of as `[]vector4` (to use Go types as an example).

</details>


# The Purpose of Broadcasting #

Is to accelerate computation. Broadcasting is an operational consideration, but has been (ab)used well for neural network work.

<details>
<summary>Click to expand</summary>

Consider broadcasting an add operation of the example of `a + b` from above. If you can view `a` as `[]vector4` then `a + b` is conceptually simply the following functions (both of which give the same answer):

```
func broadcastedAdd1(a []vector4, b vector4) (retVal []vector4) {
	for _, v := range a {
		c = vec4add(v, b)
		retVal = append(retVal, c)
	}
	return
}

func broadcastAdd2(a []vector4, b vector4) (retVal []vector4) {
	b_ := make([]vector4, len(a))
	for i := range tmp {
		copy(b_[i], b)
	}

	for i, v := range a {
		c := vec4Add(v, b_[i])
		retVal = append(retVal, c)
	}
	return retVal
}
```

Observe that the amount of memory required for this operation is simply much less than if `b` were repeated as

```
b_ =⎡10 20 30 40⎤
    ⎢10 20 30 40⎥
    ⎣10 20 30 40⎦
```

`b_` requires 3x more memory than `b`. `broadcastAdd1` uses more memory than `broadcastAdd2`. Now of course, the issue is a lot more subtle - depending on varioua SIMD instructions, it might be better to do `broadcastAdd2`. At any rate, broadcasting is an operational optimization. It requires some minor twisting of semantics of what matrices are.

This semantic twisting as it turns out is quite useful for building neural networks. The outermost dimensions are often treated as a batch of inputs. This allowed neural networks to do batch learning, leveraging the BLAS operations with much efficiency. As such, broadcasted operations are the norm.
</details>

# Early Versions of Broadcasting #

The `Broadcast` function in Gorgonia has a very long and convoluted history. The goal of Gorgonia is to provide clear semantics. Broadcasts often added hidden magic to semantics. For example, many people new to building neural networks frequently trip up on the addition of biases. The current design of `Broadcast` in Gorgonia is not good and is a sore point.

<details>
<summary>Click to expand</summary>

A neural network may be defined simply as this equation: \sigma(W'x + b)). Here, we use capital letters denote a matrix, while small letter variables denote a vector.

This works fine if the input `x` were a vector. But extend this idea to batched inputs, then the equation changes to \sigma(XW + b)). Observe that `b` is still a vector. A common mistake (even as someone who has worked on neural networks for the past 8 years) is to create a matrix for `b` (i.e. the equation is then \sigma(XW + B)). This often leads to overparameterization (although I did enjoy exploring that idea in the past few years).

The original idea of `Broadcast` was then of course to allow operations to be done in such a way that was broadcastable. However, as @owulveryck noted a few years ago, the elementary binary operations data type is internal to the package. By the time Gorgonia was released, I had already settled on a pattern of manually using `Repeat` to "broadcast". It was a fraught operation. I agreed with @owulveryck, and so added the `BroadcastXXX` functions.

But `Broadcast` had been poorly designed to begin with. Before I released Gorgonia, an even earlier version of `Broadcast` existed as a decorator. This version looked something like this:

```
func Broadcast(op BinaryOp, l, r []byte) (func(*Node, *Node) (*Node, error)) {
	return func(a, b *Node) (*Node, error) {
		// do the repeat stuff here
		a2, b2 = ...
		return ApplyOp(op, a2, b2)
	}
}
```

This was quickly modified to the current version for two reasons:

1. `Op`, `ApplyOp`, etc  wasn't originally exported.
2. I wanted there to be an opportunity to optimize any binary operation - the VMs may use privileged information about broadcasting ops to run optimized versions. An example would be like the `linAlgBinOp` type: [here](https://github.com/gorgonia/gorgonia/blob/master/op_math.go#L516)

</details>

However, now that we have this issue, we may reconsider a variant of the decorator pattern:

```
func Broadcast(op func(*Node, *Node)(*Node, error), l, r []byte)(func(*Node, *Node)(*Node, error) {
	return func(a, b *Node)(*Node, error){
		a2, b2 = ...
		return op(a2, b2)
	}
}
```

# We Should Rethink Broadcasting #

Broadcasting is here to stay. But it should stay where the semantics are clear. The technical solution to solve this PR introduces additional magic. This is not good.

## Considerations ##
<details>
<summary>Click to expand</summary>

Additionally, there are many `Broadcast` subtleties that have yet to be explored. I list them here:

* The following are examples of broadcast patterns that are inferrable by the `magic()` function listed in the technical solution above:
    * Broadcasting a vector of shape (2) to a matrix of (2, 3).
	* Broadcasting a vector of shape (3) to a matrix of (2, 3) (assuming of course, the correct broadcast pattern is provided - but it can also be super magical).
	* Broadcasting a vector of shape (2) to a 3-tensor (4, 2, 3).
	* Broadcasting a matrix of shape (2, 3) to a 3-tensor (2, 3, 4).
	* Broadcasting a matrix of shape (2, 3) to a 3-tensor (4, 2, 3).
* The `MaybeBroadcast` issue. This is an issue that was sorta solved in `golgi`. `golgi` has a set of `BroadcastXXX` functions which checks to see if the inputs are of the same shape. If they are then no broadcast operation is done. I am not a fan of this solution.
* The `MagicBroadcast` issue. We can of couse infer where to insert a `Broadcast` operation based on the input shapes. This negates the need to specify the broadcast axes. But this makes it very much like Numpy.

The way I think of this is a spectrum of userfriendliness, with tradeoffs in cognitive load and ritual. The following crude ASCII diagram explains what I mean:

```
                     more implicit
                     cognitive load
                           ^                    X
                           |                  golgi
                           |
                           |
                           |
                           |
                           |
less userfriendly          |                more userfriendly
     <--------------------------------------------->
 more principled           |                 less principled
                           |
                           |
                           |
                 ?         |
          X   +----->      |
       gorgonia            v
                     less implicit
                     cognitive load

```

The question is this: can we make Gorgonia more user friendly without sacrificing principles or cognitive load? Here are some followup questions (feel free to add more):

* Currently there are many `BroadcastXXX` functions. Are they necessary if we adopt a decorator style pattern? This itself reduces the API surface of Gorgonia - fewer builtin functions defined in the package.
* Can we clarify the semantics of broadcasting?  An example would be something like: "Broadcasting is allowed, but you have to make sure your dimensions match", so the user has to manually do the reshape, as opposed to the library doing it for them.

These of course, would invalidate a number of test cases and examples. It would be considered a breaking change in API, but I think this would be warranted.

## Consider the End User ##

Who are the end users? Gorgonia follows the Go principle - software source code is meant to be read by many. "Many" in this sense also means "diverse". I strongly suspect the people who are reading my code long after I have deployed them are not going to be machine learning engineers who know the details of Numpy's broadcasting rules (even I don't fully remember it!).

What can we leverage in this proposed group of users? If everyone treats broadcasting as the norm, then we can leverage that. Otherwise we can't, and we would need to make broadcasting explicit. I'm in favour of making broadcasting explicit, and explicit to a point that just by reading the code I know what is happening.

</details>


# Additional Notes #

While exploring this issue, I have noted several issues that sorta bugged me, which may be rolled into this pull request.

1. Rename `ExampleBroadcastAdd` to `Example_Broadcasting` - this should be a global example, of which `BroadcastAdd` was used as an example.
2. The example `ExampleBroadcastAdd` is not particularly illuminating. `a` should be a vector of `[100 200]`. This way we can see how it is added along the axes. Of course this issue is moot if we decide to be more strict on the broadcasting operations.

owulveryck:
Let me express another pain of a typical end-user of Gorgonia:
We need to wait for the graph execution to notice a "broadcast error."

While writing code with Gorgonia, the developer is assisted. Here are different verification steps:

* Go tooling (`gofmt`, `govet`, ... ) ensure that code is idiomatic
* Go compiler ensure that code is valid
* Gorgonia's API should ensure that usage is allowed
* Gorgonia's runtime (`VM`) does some extra checks on each operator.

The "broadcasting" errors are usually raised during the last step, which is too late.
If we could move it to the very top (with a proper tool would be great), the IDE would help us diagnose it. Therefore the user experience would be significantly enhanced.
To achieve this, the tool must use primitives offered by the Gorgonia API itself; therefore, it needs to be implemented in step 3.

This is, IMHO, mandatory if we want to use this:
> Can we clarify the semantics of broadcasting? An example would be something like: "Broadcasting is allowed, but you have to make sure your dimensions match", so the user has to manually do the reshape, as opposed to the library doing it for them.

chewxy:
> We need to wait for the graph execution to notice a "broadcast error."

I think this is due to a mismatch between the APIs of `gorgonia.org/gorgonia` and `gorgnia.org/tensor` - the `tensor` variant of things generally do not keep the dimensions. 

@owulveryck  If possible can you provide an example, so that we may put this in the KNOWN_ISSUES test while we deliberate on this topic

chewxy:
@jorgecarleitao I shall merge this. Can you please send me an email? chewxy [at] gmail.com with the title "GORGONIA CONTRIBUTOR" ?

PR #377, opened 2020-02-13
Doc/fixes

owulveryck:


owulveryck:
Small fixes on the doc:
* add monad as a keyword for the examples 
* clarify the Read method

PR #381, opened 2020-02-19
Adding test for activation function Mish

ma-hei:
Adding a unit test for activation function Mish.
Testing non-monotonicity property of Mish and
Mish's derivative.

### DIFF
--- a/operations_test.go
+++ b/operations_test.go
@@ -348,6 +348,54 @@ func TestGt(t *testing.T) {
 
 }
 
+func TestMisha(t *testing.T) {
+	defer runtime.GC()
+	assert := assert.New(t)
+	g := NewGraph()
+	var err error
+	var x0, x1, x2, f0, f1, f2 *Node
+	var grad0, grad1, grad2 Nodes
+
+	x0 = NewScalar(g, Float64, WithName("x0"))
+	x1 = NewScalar(g, Float64, WithName("x1"))
+	x2 = NewScalar(g, Float64, WithName("x2"))
+
+	Let(x0, -2.5)
+	Let(x1, -2.2)
+	Let(x2, 1.0)
+
+	f0 = Must(Mish(x0))
+	f1 = Must(Mish(x1))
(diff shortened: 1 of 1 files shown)


operations_test.go
```diff
@@ -348,6 +348,58 @@ func TestGt(t *testing.T) {
 
 }
 
+func GetF64 (n *Node) float64 {
```
ma-hei: Is there a simpler way to do this? I need to get the value of the node in order to compare it with assert.Greater/assert.Less.

operations_test.go
```diff
@@ -348,6 +348,58 @@ func TestGt(t *testing.T) {
...
+
+	Let(x0, -2.5)
+	Let(x1, -2.2)
+	Let(x2, 1.0)
+
+	f0 = Must(Mish(x0))
+	f1 = Must(Mish(x1))
+	f2 = Must(Mish(x2))
+
+	if grad0, err = Grad(f0, x0); err != nil {
```
ma-hei: I'm repeating myself here. Maybe there's a better way?

operations_test.go
```diff
@@ -697,6 +749,7 @@ func TestMean(t *testing.T) {
 	}
 }
 
+
```
ma-hei: I'll remove this new line.

chewxy:
Lightning storm melted my computer's power cable into motherboard. Will review later on a different machine (but browsing via phone indicates it looks good)

operations_test.go
chewxy: https://github.com/gorgonia/gorgonia/blob/master/testsetup_test.go#L53

chewxy:
@ma-hei  all good! Those two comments are minor issues. You can fix them if you want (or I can just merge this if you aren't free). More importantly, can you please email me? chewxy [at] gmail. More via email

ma-hei:
I removed the unnecessary new line and used the existing "extractF64" method. Feel free to merge! Sure, I'll send an email later today!

PR #388, opened 2020-03-25
Fix Dropout() to return corrrect scaled output

### LINKED ISSUE #387: Bug in Dropout function
Hi people,

I have a model that uses dropout during training, I'm using the exact same model but without the dropout layers during validation. For the validation step I copy all the learned values over to the validation model and run it over the validation data.

Expectation: As the training cost goes down the validation cost will also decrease.
```
cost 107.09594 | acc 0.03626 | val_cost 106.57639 | val_acc 0.00000 
cost 106.46517 | acc 0.06484 | val_cost 106.11518 | val_acc 0.05714 
cost 102.56391 | acc 0.08462 | val_cost 100.13887 | val_acc 0.20000 
cost 98.47047 | acc 0.11978 | val_cost 98.32637 | val_acc 0.20000 
cost 95.26812 | acc 0.14945 | val_cost 97.95923 | val_acc 0.25714
```
Actual: The validation cost skyrockets.
```
cost 106.56745 | acc 0.03187 | val_cost 103.83164 | val_acc 0.00000 
cost 106.14746 | acc 0.04725 | val_cost 137.27169 | val_acc 0.02857 
cost 104.14275 | acc 0.05604 | val_cost 345.20932 | val_acc 0.08571 
cost 99.59372 | acc 0.10440 | val_cost 423.45196 | val_acc 0.05714 
cost 97.69338 | acc 0.10000 | val_cost 288.48028 | val_acc 0.17143 
```

If I also enable dropout during validation the cost decreases like expected. So I think the copying of the learned parameters works. Although maybe not optimal:
```go
learnablesTrain := mTrain.Learnables()
learnablesTest := mTest.Learnables()
for i := range learnablesTrain {
	if err = gorgonia.Let(learnablesTest[i], learnablesTrain[i].Value()); err != nil {
		return err
	}
}
```

---

Looking at the current dropout implementation it looks like it tries to implement inverted dropout. So the values during training are increased to compensate for the decreased number of connections. This is in contrast to normal dropout where the outputs are decreased during inference. Assuming this is true, then I do not have to add any compensation myself in my validation model.

[Current dropout implementation](https://github.com/gorgonia/gorgonia/blob/39b728c1ca24b82cee00c838bed7d318c6bcbd2a/nn.go#L65):
```go
// Dropout is a convenience function to implement dropout.
// It uses randomly zeroes out a *Tensor with a probability drawn from
// a uniform distribution
func Dropout(x *Node, prob float64) (retVal *Node, err error) {
	if prob == 0.0 {
		return x, nil
	}

	var dt tensor.Dtype
	if dt, err = dtypeOf(x.t); err != nil {
		return nil, errors.Wrap(err, dtypeOfFail)
	}

	var opp, pr Value // opp = 1 per p
	switch dt {
	case Float64:
		opp, _ = anyToScalar(1.0 / prob)
		pr, _ = anyToScalar(prob)
	case Float32:
		opp, _ = anyToScalar(float32(1.0 / prob))
		pr, _ = anyToScalar(float32(prob))
	default:
		return nil, errors.Errorf(nyiTypeFail, "Dropout()", dt)
	}

	p := NewConstant(pr)
	c := NewConstant(opp)

	m := UniformRandomNode(x.g, dt, 0, 1, x.shape...)
	if retVal, err = Gt(m, p, true); err != nil {
		return nil, errors.Wrap(err, "Greater Than failed")
	}

	if retVal, err = HadamardProd(x, retVal); err != nil {
		return nil, errors.Wrap(err, mulFail)
	}

	return HadamardDiv(retVal, c)
}
```
I expect the last line to *increase* the return value. But `c = 1.0 / prob` which makes the whole thing `return retVal / (1.0 / prob)` <=> `return retVal * prob`, *decreasing* it instead.

Proposed fix:
```go
// Dropout is a convenience function to implement dropout.
// It uses randomly zeroes out a *Tensor with a probability drawn from
// a uniform distribution
func Dropout(x *Node, prob float64) (retVal *Node, err error) {
	if prob == 0.0 {
		return x, nil
	}

	var dt tensor.Dtype
	if dt, err = dtypeOf(x.t); err != nil {
		return nil, errors.Wrap(err, dtypeOfFail)
	}

	var pr Value // opp = 1 per p
	switch dt {
	case Float64:
		pr, _ = anyToScalar(prob)
	case Float32:
		pr, _ = anyToScalar(float32(prob))
	default:
		return nil, errors.Errorf(nyiTypeFail, "Dropout()", dt)
	}

	p := NewConstant(pr)

	m := UniformRandomNode(x.g, dt, 0, 1, x.shape...)
	if retVal, err = Gt(m, p, true); err != nil {
		return nil, errors.Wrap(err, "Greater Than failed")
	}

	if retVal, err = HadamardProd(x, retVal); err != nil {
		return nil, errors.Wrap(err, mulFail)
	}

	return HadamardDiv(retVal, p)
}
```
This gives me the expected result when I run my prediction model without dropout.

MarkKremer:
Resolves #387 

I also extracted the `Dropout` function into a private `dropout` function so I can inject my own "random" numbers during tests.

Let me know if I need to change anything. :+1:

### DIFF
--- a/nn.go
+++ b/nn.go
@@ -63,6 +63,12 @@ func BinaryXent(output, target *Node) (retVal *Node, err error) {
 // It uses randomly zeroes out a *Tensor with a probability drawn from
 // a uniform distribution
 func Dropout(x *Node, prob float64) (retVal *Node, err error) {
+	return dropout(x, prob, UniformRandomNode)
+}
+
+type dropoutRandFn func(g *ExprGraph, dt tensor.Dtype, low, high float64, shape ...int) *Node
+
+func dropout(x *Node, prob float64, randFn dropoutRandFn) (retVal *Node, err error) {
 	if prob == 0.0 {
 		return x, nil
 	}
@@ -72,22 +78,19 @@ func Dropout(x *Node, prob float64) (retVal *Node, err error) {
 		return nil, errors.Wrap(err, dtypeOfFail)
 	}
 
-	var opp, pr Value // opp = 1 per p
+	var pr Value
 	switch dt {
 	case Float64:
-		opp, _ = anyToScalar(1.0 / prob)
 		pr, _ = anyToScalar(prob)
 	case Float32:
-		opp, _ = anyToScalar(float32(1.0 / prob))
 		pr, _ = anyToScalar(float32(prob))
 	default:
 		return nil, errors.Errorf(nyiTypeFail, "Dropout()", dt)
 	}
 
 	p := NewConstant(pr)
-	c := NewConstant(opp)
 
-	m := UniformRandomNode(x.g, dt, 0, 1, x.shape...)
+	m := randFn(x.g, dt, 0, 1, x.shape...)
 	if retVal, err = Gt(m, p, true); err != nil {
 		return nil, errors.Wrap(err, "Greater Than failed")
 	}
@@ -96,7 +99,7 @@ func Dropout(x *Node, prob float64) (retVal *Node, err error) {
 		return nil, errors.Wrap(err, mulFail)
 	}
 
-	return HadamardDiv(retVal, c)
+	return HadamardDiv(retVal, p)
 }
 
 // LeakyRelu returns a node whose underlying value is:
--- a/nn_test.go
+++ b/nn_test.go
@@ -1,6 +1,7 @@
 package gorgonia
 
 import (
+	"fmt"
 	"io/ioutil"
 	"runtime"
 	"testing"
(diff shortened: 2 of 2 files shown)


MarkKremer:
I found that `dropout_cuda.go` already has a struct with the name `dropout` so there will be a conflict. I'm now installing the cuda drivers to test with that. Will fix that afterwards.

MarkKremer:
`dropout_cuda.go` is in a different package so there are no conflicts like I thought earlier. I can build the package with the `cuda` tag. Tests with cuda do fail, but they also fail on the master branch for me. I think something in my installation is still not correct.

chewxy:
Yeah CUDA is quite finnicky

chewxy:
This LGTM if you want me to merge (i.e happy with leaving the cuda things behind) let me know

MarkKremer:
I'm happy with merging. If you want you could try to run the tests with cuda on the PR but I think I'll leave it for another time.

PR #398, opened 2020-04-20
Fix download

mattn:
Collect path is `testdir/mnist/*`.


chewxy:
Can @owulveryck and @mattn handle this?

owulveryck:
I’ll merge it today.

Issue #404, opened 2020-05-30
Calculate dot product 2 times in different directions


MarkKremer:
Hey people. :wave: 

I'm trying to calculate the dot product of a matrix + vector. I have to do this 2 times in different directions (so the matrix is used as input for 2 different operations in the graph). Everything is batched and differentiation is needed. I've tried different solutions using `BatchedMatMul` and `Broadcast`->`HadamardProd`->`Sum` but I'm running into difficulties with each of them.

Here's an example using BatchedMatMul:
```golang
package main

import (
	"fmt"
	"gorgonia.org/gorgonia"
	"gorgonia.org/tensor"
	"log"
)

func main() {
	g := gorgonia.NewGraph()

	m := gorgonia.NewTensor(g, tensor.Float64, 3, gorgonia.WithShape(1, 3, 4), gorgonia.WithName("m"), gorgonia.WithValue(tensor.New(tensor.WithShape(1, 3, 4), tensor.WithBacking([]float64{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12}))))
	fmt.Printf("Matrix =\n%v", m.Value())
	// Matrix =
	//   ⎡ 1   2   3   4⎤
	//   ⎢ 5   6   7   8⎥
	//   ⎣ 9  10  11  12⎦

	a := gorgonia.NewTensor(g, tensor.Float64, 3, gorgonia.WithShape(1, 4, 1), gorgonia.WithName("a"), gorgonia.WithValue(tensor.New(tensor.WithShape(1, 4, 1), tensor.WithBacking([]float64{1, 2, 3, 4}))))
	fmt.Printf("a =\n%v", a.Value())
	// a =
	//   ⎡1⎤
	//   ⎢2⎥
	//   ⎣3⎥
	//   ⎢4⎦

	b := gorgonia.NewTensor(g, tensor.Float64, 3, gorgonia.WithShape(1, 1, 3), gorgonia.WithName("b"), gorgonia.WithValue(tensor.New(tensor.WithShape(1, 1, 3), tensor.WithBacking([]float64{1, 2, 3}))))
	fmt.Printf("b =\n%v", b.Value())
	// b =
	//   ⎡1  2  3⎤

	dot1, err := gorgonia.BatchedMatMul(m, a)
	if err != nil {
		log.Fatal(err)
	}

	dot2, err := gorgonia.BatchedMatMul(b, m)
	if err != nil {
		log.Fatal(err)
	}

[...]

PR #405, opened 2020-05-31
Copy input shape for batchedMatMulOp instead of append

MarkKremer:
If you look a bit up in the diff of `op_math.go` you'll see `outerX := x[:len(x)-2]`. Taking a range from a slice and then appending to it will override the original slice instead of copying it into a new slice. Not very obvious but that's how it works. This change does a copy so that the shape of the input node doesn't get altered.

I looked into `slice.Clone()` but returning a slice to the pool that I appended to didn't feel right. Manually calling `BorrowInts()` creates a lot of clutter. It would be nice if there was a function to clone+append in the tensor package to do this. But for now I chose this approach. However, it isn't clear if `outerX` can contain more than 1 element. If it doesn't, then I could simplify it some more.

This is a result of #404 but don't close it yet. There are still some other issues I'm having that I'm still figuring out.


MarkKremer:
#391 fixes this too. I propose I remove my fix from this branch but keep the test. Then you'll have some extra tests for your own fix. :)

chewxy:
agreed.. let me do some finnacking around the gomod things. (#391 relies on a newer version of `tensor`)

chewxy:
Will merge this into master once CI says go

PR #409, opened 2020-06-09
Make reductionInferShape conservative to fix #384

### LINKED ISSUE #384: Optimisation in NewUniqueNode may be buggy
I think this is related to #373  and #375 

I made a simple test, as described in #375:

```go
package gorgonia_test

import (
	"testing"

	"gorgonia.org/gorgonia"
)

func TestMean_issue375(t *testing.T) {
	g := gorgonia.NewGraph()

	w0 := gorgonia.NewTensor(g, gorgonia.Float32, 4, gorgonia.WithShape(1, 64, 1, 64), gorgonia.WithName("w0"), gorgonia.WithInit(gorgonia.GlorotN(1.0)))
	result, _ := gorgonia.Mean(w0, 3)
	t.Log(result)
}
```

After some investigation, the code is panicking because of a call to `WithShape` here: https://github.com/gorgonia/gorgonia/blob/2d4605d582ad6f24cd62ee7860e907d828fbe28c/node.go#L210-L212

which is called here: https://github.com/gorgonia/gorgonia/blob/2d4605d582ad6f24cd62ee7860e907d828fbe28c/op.go#L202

The problem is in the function `NewUniqueNode` which calls `newNode`... which calls `n.borrowNode` for optimization: https://github.com/gorgonia/gorgonia/blob/2d4605d582ad6f24cd62ee7860e907d828fbe28c/node.go#L259

It applies the `WithShape` function here ...https://github.com/gorgonia/gorgonia/blob/2d4605d582ad6f24cd62ee7860e907d828fbe28c/node.go#L264 
therefore, all the checks are performed on the node that is borrowed and full of rubbish!

In the example, there is only one node to be borrowed; therefore, the panic always returns the same result, but with a more significant code, the problem could appear sporadically.

wzzhu:
The reductionInferShape currently doesn't respect along initially.
It aggressively squeezes dimensions. Not only does it affect normal tensor operation, but also it breaks the backprop autoDiff algorithm sometimes when the network containing BroadcastAdd, resulting in crash when calling Grad().

The change tries to strictly respect the parameter along, e.g.,
(100, 1) along 0, reduction to shape (1) instead ()
(1, 64, 1, 64) along 3 will reduce to (1, 64, 1)
(64, 1, 3, 2) along (2,3) will reduce to (64, 1).

### DIFF
--- a/differentiation.go
+++ b/differentiation.go
@@ -31,7 +31,7 @@ func forwardDiffAnalysis(outputs, sortedNodes Nodes) (retVal NodeSet, err error)
 	// diffSet := outputs.Set()
 	diffSet := outputs.mapSet()
 
-	symdiffLogf("Diff Set: %d", diffSet)
+	symdiffLogf("Diff Set: %v", diffSet)
 	symdiffLogf("%d", sortedNodes)
 	// for i := len(sortedNodes) - 1; i ⩾ 0; i-- {
 	// 	n := sortedNodes[i]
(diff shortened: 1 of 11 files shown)


chewxy:
Will look into this in the arvo.

wzzhu:
Thanks. The change breaks some unit tests. Will check it and re-submit new updates.

wzzhu:
resubmit a cleaner one with one commit, remove this one first.

PR #411, opened 2020-06-11
Make reductionInferShape conservative to fix gorgonia#384

wzzhu:
The reductionInferShape currently doesn't respect along initially.
It aggressively squeezes dimensions. Not only does it affect normal tensor operation, but also it breaks the backprop autoDiff algorithm sometimes when the network containing BroadcastAdd, resulting in crash when calling Grad().

The change tries to strictly respect the parameter along, e.g.,
(100, 1) along 0, reduction to shape (1) instead ()
(1, 64, 1, 64) along 3 will reduce to (1, 64, 1)
(64, 1, 3, 2) along (2,3) will reduce to (64, 1).

Fixed unit tests.

### DIFF
--- a/differentiation.go
+++ b/differentiation.go
@@ -31,7 +31,7 @@ func forwardDiffAnalysis(outputs, sortedNodes Nodes) (retVal NodeSet, err error)
 	// diffSet := outputs.Set()
 	diffSet := outputs.mapSet()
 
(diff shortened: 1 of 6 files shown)


chewxy:
LGTM. Sorry for the late review. Life has been nuts and I barely have time at the computer

PR #419, opened 2020-07-07
feat: attempt for a github action

owulveryck:
~This action test with GOPATH on Linux amd64~ This does not work 
Then it test with gomodules on linux / darwin / amd64 / arm

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -5,16 +5,19 @@ env:
 jobs:
   test-gopath:
     env:
-      GOPATH: ${{ github.workspace }}
       GO111MODULE: off
     strategy:
       matrix:
         go: [1.14]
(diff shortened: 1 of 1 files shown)


owulveryck:
I propose this github action.
First it runs test for go 1.13. and go 1.14 and coverage for go 1.14 in parallel.
If tests are success, it runs a build on amd64 Linux and macos in parralel.

All the steps are completing before travis :D 

WDYT?

blackrez:
Why use github actions instead of travis ?
Also you only test build on linux/amd64 because github actions doesn't support other architecture.

blackrez:
We could add in the future a arm runner with hosted runner. Looks good to me https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners

PR #423, opened 2020-07-21
Gomachine/refacto

owulveryck:
This is a refacto of the GoMachine.

The API is not stable yet (it may change when adding autodiff).
It needs some testing, and any test is welcome.

This is a bench comparison that looks promising, that computes a trivial Conv2d operator (the comparison is made against the `TapeMachine` :

```
❯ benchcmp /tmp/2 /tmp/1
benchmark                  old ns/op     new ns/op     delta
BenchmarkMachine_Run-4     172056        119103        -30.78%

benchmark                  old allocs     new allocs     delta
BenchmarkMachine_Run-4     652            219            -66.41%

benchmark                  old bytes     new bytes     delta
BenchmarkMachine_Run-4     23870         12261         -48.63%
```


owulveryck:
I made a lot of fixes.
I was not waiting for all goroutines to finish, this was an issue when running multiple times (this was causing the benchmark to fail).
The result is that it is now slower than the tape machine.

Here is a benchmark on the tiny-yolo-coco:

```
❯ go test -bench=. -benchmem -benchtime=30s
goos: darwin
goarch: amd64
pkg: gorgonia.org/gorgonia/examples/tiny-yolo-v2-coco
BenchmarkGoMachine-4                  61         742574409 ns/op        314103253 B/op      1481 allocs/op
BenchmarkTapeMachine-4                54         639415245 ns/op        147627449 B/op      1717 allocs/op
PASS
ok      gorgonia.org/gorgonia/examples/tiny-yolo-v2-coco        111.713s
```

```go
package main

import (
	"context"
	"testing"

	"gorgonia.org/gorgonia"
	G "gorgonia.org/gorgonia"
	xvm "gorgonia.org/gorgonia/x/vm"
	"gorgonia.org/tensor"
)

func BenchmarkGoMachine(b *testing.B) {
	// Init Graph
	g := G.NewGraph()

	model := NewTinyYOLOv2Net(g, 80, 5, "model/yolov2-tiny.weights")

	imgf32, err := GetFloat32Image("data/dog_416x416.jpg")
	if err != nil {
		b.Fatal(err)
	}

	image := tensor.New(tensor.WithShape(1, channels, height, width), tensor.Of(tensor.Float32), tensor.WithBacking(imgf32))
	x := gorgonia.NewTensor(g, tensor.Float32, 4, gorgonia.WithShape(1, channels, width, height), gorgonia.WithName("x"))

	gorgonia.Let(x, image)
	if err := model.FeedForward(g, x); err != nil {
		b.Fatalf("%+v", err)
	}

	machine := xvm.NewMachine(g)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		err = machine.Run(context.Background())
		if err != nil {
			b.Fatal(err)
		}
	}
	machine.Close()
}

func BenchmarkTapeMachine(b *testing.B) {
	// Init Graph
	g := G.NewGraph()

	model := NewTinyYOLOv2Net(g, 80, 5, "model/yolov2-tiny.weights")

	imgf32, err := GetFloat32Image("data/dog_416x416.jpg")
	if err != nil {
		b.Fatal(err)
	}

	image := tensor.New(tensor.WithShape(1, channels, height, width), tensor.Of(tensor.Float32), tensor.WithBacking(imgf32))
	x := gorgonia.NewTensor(g, tensor.Float32, 4, gorgonia.WithShape(1, channels, width, height), gorgonia.WithName("x"))

	gorgonia.Let(x, image)
	if err := model.FeedForward(g, x); err != nil {
		b.Fatalf("%+v", err)
	}

	machine := gorgonia.NewTapeMachine(g)
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		err = machine.RunAll()
		if err != nil {
			b.Fatal(err)
		}
		machine.Reset()
	}
	machine.Close()
}
```

owulveryck:
Profiling the benchmark gives this:
Screenshot 2020-07-22 at 12 37 58
This sounds normal because we are triggering *a lot* of goroutines.

I should try on a machine with more CPU to see how it behaves.

An optimization idea would be to trigger the goroutines in a smart way (not all of them at the same time), by defining stages inside the machine.

something like:

```go
type Machine struct {
     stages [][]*node
    // ...
}

func (m *Machine) Run(context.Context) error {
    for i := range m.stages {
        for j := range m.stages[i] {
            go run m.stages[i][j].run()
        }
    }
}

**Comment by MarkKremer** [comment]
My 2 cents (I haven't read a lot of what you've done in this PR because I'm on my phone):

I think stages can be used for optimization but it isn't the optimal solution. For example, if you have 3 short task in series and 1 long task parallel to those 3. In what stage would you put the long task? Unnecessary waiting would happen.

**Comment by owulveryck** [comment]
@MarkKremer I agree, the solution is not optimal, but the stages would not be composed of "short" or "long" running processes.
A stage would be a group of noded at the same depth regarding the bottom of the expr graph.

Take for example the following formula: `(a + b) * (c + d)`

that would lead to this graph:

```dot
a -> op1(+)
b -> op1(+)
c -> op2(+)
d -> op2(+)
op1(+) -> op(*)
op2(+) -> op(*)
```

We could have three groups:

- group1: a,b,c,d
- group2: op1(+),op2(+)
- group3: op(*)

On a simple graph, this is easy to see the benefit. 
But, as each not is trigger within a go-routine, on larger graphs, we must carefully build the groups to avoid deadlocks. Eventually, we need to be smarter than go's internal scheduler. I guess that we can leave it as it is for now :D

Issue #435, opened 2020-10-01
MaxPool1D: Impossible height/kernel/pad combination


strigi-form:
MaxPool1D(x *Node, 2, 0, 1)
with x being a vector (node from Dense with single dim []int{n}) returns the error:
errors.New("Impossible height/kernel/pad combination")

My guess is that pad/size checks are done with 2D tensor in mind.

I'm proposing a fix

chewxy:
Fixed by #436 I believe

PR #436, opened 2020-10-01
MaxPool2D checks (#435)

strigi-form:
Check all padding (N S E W), kernel and input size to avoid index out of range


chewxy:
LGTM. Thanks for the fix. Can you send me an email chewxy@gmail.com with the title "Gorgonia Contributor"? Thanks

PR #437, opened 2020-10-07
Write sparsemax operation

dcu:


chewxy:
LGTM. Might wanna add a note that currently it only supports float64

PR #440, opened 2020-10-10
Softmax op

dcu:


op_softmax.go
```diff
@@ -0,0 +1,272 @@
...
+		return nil, fmt.Errorf("Can't check SoftmaxDiff input: %w", err)
+	}
+
+	diag := tensor.New(tensor.AsDenseDiag(inputTensor))
+
+	sm := inputTensor.Clone().(tensor.Tensor)
+	sm.Reshape(inputTensor.Shape().TotalSize(), 1)
+
+	smT := sm.Clone().(tensor.Tensor)
+	smT.Transpose()
```
chewxy, line 250: you can use `smT.T()`, which does no work, rather than `Transpose` which does a lot of work

op_softmax.go
```diff
@@ -0,0 +1,272 @@
...
+
+	smDot, err := tensor.Dot(sm, smT)
```
chewxy, line 252: Use `MatMul`.  `tensor.Dot` is a bad habit

chewxy:
Phew. This LGTM. Once it passes tests, you can merge. Most excellent work @dcu

PR #453, opened 2020-11-04
Check Let inputs

dcu:
The shapes should be the same

### DIFF
--- a/gorgonia.go
+++ b/gorgonia.go
@@ -266,6 +266,10 @@ func UnsafeLet(n *Node, be interface{}) error {
 		}
 
 	case Value:
+		if !n.Shape().Eq(v.Shape()) {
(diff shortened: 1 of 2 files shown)


gorgonia.go
```diff
@@ -266,6 +266,10 @@ func UnsafeLet(n *Node, be interface{}) error {
 		}
 
 	case Value:
+		if !n.Shape().Eq(v.Shape()) {
+			return fmt.Errorf("Unable to let %v. Expected Shapes to be equal. Got %v and %v instead", n.name, n.Shape(), v.Shape())
```
chewxy, line 270: Wondering if you could change the error msg to be clearer? Specifically the last sentence: "Node's expected shape is %v. Got %v instead"

dcu: done, tests are running

dcu:
all done

PR #455, opened 2020-11-16
Attempt to reproduce a reshape bug

chewxy:


chewxy:
My reaction to this as I am reading this for the first time: WTF?

chewxy:
Not sure if this is really a bug. I understand the issue now.  It happens in the back propagation phases. 

The error is as such:

```
 Input shape is (100, 3136). Expected (100, 64, 7, 7)
```
They're the same exact sizes (64 * 7 * 7 = 3136). This may be an overzealous checker

chewxy:
line 1159 in `op_tensor.go`. I gotta rethink this for a bit.

dcu:
wouldn't checking the total size do the trick?

chewxy:
I believe #457 fixes this

PR #459, opened 2020-12-13
Make sparsemax an axial operation

dcu:
And support any dimension for the input

### DIFF
--- a/op_sparsemax.go
+++ b/op_sparsemax.go
@@ -11,17 +11,26 @@ import (
 )
 
 type sparsemaxOp struct {
+	axis int
 }
 
-func newSparsemaxOp() *sparsemaxOp {
-	sparsemaxop := &sparsemaxOp{}
+func newSparsemaxOp(axes ...int) *sparsemaxOp {
+	axis := -1
+	if len(axes) > 0 {
+		axis = axes[0]
+	}
+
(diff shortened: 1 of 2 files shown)


op_sparsemax.go
```diff
@@ -94,7 +116,7 @@ func (op *sparsemaxOp) Do(inputs ...Value) (Value, error) {
 		return nil, fmt.Errorf("invalid input type for Sparsemax, expected float64 or float32, got: %v", inputTensor.Dtype())
 	}
 
-	return tensor.New(tensor.Of(inputTensor.Dtype()), tensor.WithShape(inputTensor.Size()), tensor.WithEngine(inputTensor.Engine()), tensor.WithBacking(output)), nil
+	return tensor.New(tensor.Of(inputTensor.Dtype()), tensor.WithShape(inputTensor.Shape()...), tensor.WithEngine(inputTensor.Engine()), tensor.WithBacking(output)), nil
```
chewxy, line 119: `.Shape().Clone()...` - this will prevent funny shared shape things from happening

chewxy (changes requested):
asides ffrom that one shape thing, all good

op_sparsemax.go
dcu: fixed

PR #463, opened 2021-01-10
Add by indices op and improve sparsemax

dcu:
- new ByIndices op
- new BatchNorm1d op
- sparse max fixes


chewxy:
This good to go? @dcu

dcu:
yes, it should be ready now.
I just pushed some local changes that I had, including BatchNorm1d similar to https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm1d.html

Issue #466, opened 2021-01-23
TapeMachine's behaviour is not idempotent after a call to Reset in some circumstances


owulveryck:
In this use-case https://github.com/gorgonia/agogo/issues/4 two consecutive execution of the TapeMachine with a call to Reset does not provide the same execution result.

It looks like the `Reset` function should reset the `cpumem` registries like this:

```diff
 // Reset resets the run state of the machine by changing the instruction pointer back to 0
+// and reseting the registry
 func (m *tapeMachine) Reset() {
        m.pc = 0
        m.ExternMetadata.Reset()
@@ -116,6 +117,9 @@ func (m *tapeMachine) Reset() {
                returnValue(m.gpumem[i])
                m.gpumem[i] = nil //
        }
+       for i := range m.cpumem {
+               m.cpumem[i] = nil
+       }
 }
```

I suspect that this instruction https://github.com/gorgonia/gorgonia/blob/12e31c4befd99d1c0844d0d7555b102c3aa11639/vm_tape.go#L509-L512

from the method https://github.com/gorgonia/gorgonia/blob/12e31c4befd99d1c0844d0d7555b102c3aa11639/vm_tape.go#L499 is causing the issue.

A test should validate the hypothesis.

chewxy:
let's merge this

dcu:
ok

PR #467, opened 2021-01-23
fix: reset the registries

owulveryck:


owulveryck:
This should fix #466

dcu:
this fixes and issue I was struggling with, thanks Olivier

dcu:
is it possible to write a test for this ?

owulveryck:
> is it possible to write a test for this ?

I agree with you, I'd like to write a test before actually merging the PR.
It is easy to reproduce with agogo  (see https://github.com/gorgonia/agogo/blob/ce6073931e2c36f2986cd1048fc89ceb80c7df7a/dualnet/dual_test.go#L17) but the graph is big, and it is not easy to find the root cause.
I am still investigating.

chewxy:
This is straightforwards and correct IMO.

PR #468, opened 2021-01-24
fix: GitHub action's new policy and housekeeping in the Go versions

owulveryck:


chewxy:
what is the new policy? Why are we testing fewer versions now?

owulveryck:
GitHub has removed the execution of `::set-env` and `::add-path::` recently. 

They documented it here: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

Regarding the execution matrix, as Gorgonia supports the last two stable versions of Go (like gonum), I think we can drop 1.13x to spare some time in the CI.

chewxy:
OK. LMK if I can merge

PR #470, opened 2021-02-05
Fix panic message

bandreghetti:
Just a simple fix to a wrong panic message I stumbled upon.


chewxy:
Nice! Thanks! Will merge when it passes. Ping me if I forget.

PR #472, opened 2021-02-10
fix(concurrent example): process partial filled chunks too

chrmang:


example_concurrent_training_test.go
```diff
@@ -149,7 +149,7 @@ func prep() (x, y Value, bs []batch) {
 }
 
 func concurrentTraining(xV, yV Value, bs []batch, es int) {
-	threads := runtime.GOMAXPROCS(-1) // reserve one thread for the CPU locked thread
```
chrmang, line 152: The comment was misleading. `-1` doesn't change the current setting of max cpus.
https://golang.org/pkg/runtime/#GOMAXPROCS says, this call will go away, when scheduler improves, so I switched to NumCPU()

example_concurrent_training_test.go
```diff
@@ -80,7 +80,7 @@ func (t *concurrentTrainer) train(x, y Value, costChan chan cost, wg *sync.WaitG
 func trainEpoch(bs []batch, ts []*concurrentTrainer, threads int) {
 	// costs := make([]float64, 0, len(bs))
 	chunks := len(bs) / len(ts)
-	for chunk := 0; chunk < chunks; chunk++ {
```
chrmang, line 83: It's a integer division. Due to the remainder we need a '<='.

example_concurrent_training_test.go
owulveryck: Nice catch !

owulveryck (approved):

PR #473, opened 2021-02-11
chore(ci): New layout of the CI

owulveryck:
This is a complete rewrite of the CI factory.

## Workflows

There are now three workflows corresponding to the different OS of the runners:
- Build and Tests on Linux/amd64 
- Build and Tests on MacOS/amd64 
- Build and Tests on Self-Hosted (arm)

## Jobs

Each workflow is composed of two jobs:
- stable-go
- previous-go

Both jobs are identical besides the go version they are executing. The reason to avoid the `matrix` strategy is that GitHub appends the matrix tags to each of the running jobs. As a consequence, changing the version of Go implies modifying the protection policy of the project.

### Caching
Each job is trying to use the cache mechanism to lower the time passed in the CI.

### Steps
The jobs are:

- Building the code for the current platform
- cross-compiling the code for any other platform 
- Launching some tests with the race condition

A matrix strategy is only present to tests the build tags (as it was before). A special tag "none" is present. If `tag` is none, the `-tag` is omitted.

## README

With this build process, I think it is safe to use the badge of the Linux runner on the README. It performs all the tests on Linux and validates that the code is compilable on all the platforms.

## TODO

Once this PR is validated, the protection of the master branch needs to be tweaked to request the validation of:

- Build and Tests on Linux/amd64 / Build and test on latest stable Go release (none)
- Build and Tests on Linux/amd64 / Build and test on latest stable Go release (avx) **if mandatory**
- Build and Tests on Linux/amd64 / Build and test on latest stable Go release (see) **if mandatory**
- ...:

## Maintenance

When a new Go version is dropped, there is a need to modify the three workflows

## Optimization

The code is heavily repeated (DRY they said). If this system is ok for us, it could be good to write a small tool (based on templates), to automate all the stuff when a new Go version is dropped.


owulveryck:
I forgot to add the coverage in the CI.
I will do that soon

owulveryck:
Good to merge if the review is ok.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.15.x
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Calc coverage
+      run: |
+        export PATH=$PATH:$(go env GOPATH)/bin
```
chrmang, line 15: Is this really necessary? If `go` is not in `PATH`, `go env` would also fail.

.github/workflows/job-template.go
```diff
@@ -0,0 +1,107 @@
...
+type job struct {
+	JobID     string          //stable-go
+	JobName   string          //Build and test on latest stable Go release
+	RunsOn    string          //ubuntu-latest
+	GoVersion string          //1.15.x
+	Tags      map[string]bool // none:false, avx:true, sse:true
+	WithRace  bool            // false
+}
+
+const workflowTmpl = `## DO NOT EDIT - This file is generated
```
chrmang, line 24: The template is a good idea, to keep GitHub workflows consistent.

chrmang (approved):

.github/workflows/coverage.yml
owulveryck: You are right; I just copied what was present before; I will remove this.
thank you

chewxy:
let's make AVX and SSE2 mandatory

owulveryck:
Can I merge? (and take care of the settings of the project to make things mandatory)

PR #479, opened 2021-02-23
fix(tanhdiff): add explicit cast for tanh test

owulveryck:
The spec mention that the behavior can differ accross different
architecture.
See https://github.com/golang/go/issues/44528 for more details


owulveryck:
Dropping it in favor of #480

PR #480, opened 2021-02-23
Fix arm64

dcu:
This fixes all of arm64 broken tests and behaviors.

### DIFF
--- a/operatorPointwise_unary_test.go
+++ b/operatorPointwise_unary_test.go
@@ -431,11 +431,11 @@ func TestTanhDiff(t *testing.T) {
 	}
 
 	correct := 1.0 - (float64(math.Tanh(v)) * float64(math.Tanh(v))) // I'm surprised Golang doesn't have a secant function!
-	assert.Equal(correct, x.boundTo.(*dualValue).d.Data())
+	assert.InDeltaf(correct, x.boundTo.(*dualValue).d.Data(), 1e-14, "")
 
 	// Tensor edition
 	xdvd := xT.boundTo.(*dualValue).d.(*tensor.Dense)
-	assert.Equal([]float64{correct, correct}, xdvd.Data())
+	assert.InDeltaSlicef([]float64{correct, correct}, xdvd.Data(), 1e-14, "")
 }
 
 func TestSigmoidDiff(t *testing.T) {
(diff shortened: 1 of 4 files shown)


operatorPointwise_unary_test.go
```diff
@@ -430,12 +430,12 @@ func TestTanhDiff(t *testing.T) {
...
 
-	correct := 1.0 - (math.Tanh(v) * math.Tanh(v)) // I'm surprised Golang doesn't have a secant function!
-	assert.Equal(correct, x.boundTo.(*dualValue).d.Data())
+	correct := 1.0 - (float64(math.Tanh(v)) * float64(math.Tanh(v))) // I'm surprised Golang doesn't have a secant function!
+	assert.InDeltaf(correct, x.boundTo.(*dualValue).d.Data(), 1e-14, "")
 
 	// Tensor edition
 	xdvd := xT.boundTo.(*dualValue).d.(*tensor.Dense)
-	assert.Equal([]float64{correct, correct}, xdvd.Data())
+	assert.InDeltaSlicef([]float64{correct, correct}, xdvd.Data(), 1e-14, "")
```
owulveryck, line 438: This is good enough for the test.
However, if we don't find a proper fix(that validates the equal asertions), we may eventually face another issue that is related. WDYT?

dcu: I think a difference of 1e-14 should not significant, but let's see what @chewxy thinks about this

solvers_test.go
```diff
@@ -362,7 +362,7 @@ func TestBarzilaiBorweinSolver(t *testing.T) {
 
 	solver := NewBarzilaiBorweinSolver(WithLearnRate(0.0001))
 
-	maxIterations := 200
+	maxIterations := 250
```
dcu, line 365: amd64 takes 198 iterations and arm64 takes 210 to reach the expected min cost

operatorPointwise_unary_test.go
chewxy: Can you add a comment on why this is used (i.e. no guarantees of identical behaviours across architecture) - link to https://github.com/golang/go/issues/44528#issuecomment-783975492

dcu: done

PR #482, opened 2021-03-09
Remove BatchNorm1d function

dcu:
It can be done with just one function for both 1d and 2d batch norms.


chewxy:
I love this. Smaller API surface. What a brilliant thing

PR #483, opened 2021-03-27
Implement average pool 1d and 2d operations

dcu:
This branch also includes 2 bugfixes for maxpool:

- crash when using a shape without checking its dims first
- crash when the calculated strides are empty


chewxy:
LGTM. I'll remove the `log` things and merge

PR #484, opened 2021-04-18
Batchnormimpl2

chewxy:


dcu:
I compared the output against pytorch's and it matches, the gradients also match but the variance do not.
I added tests to verify this and looked at the code but couldn't find the issue

Also, since the evaluation require the mean and variance I think we need to expose both so they can be saved and reloaded later otherwise it wouldn't work I believe

dcu:
I made a few changes to the variance and inference calculation. I verified everything against the pytorch's output for this operation

dcu:
actually there's still something wrong with the scale gradient

dcu:
https://github.com/gorgonia/gorgonia/pull/504 replaces this PR so I'm closing this one

PR #491, opened 2021-06-14
Adding the concept of training/eval mode to the VM

dcu:
I'm submitting this to start the discussion and understand if this feasible
Without this I think the only option to disable the Dropout op in eval mode,
for example, is recreating the graph which is hard when you want to create a
higher level framework and still let people use gorgonia directly


chewxy:
I'm back. Still sick and sinuses still heavily infected but I think this is somewhat a good idea though I think the interface should be one method:

```
type EvalModeOp interface{
    SetTraining(isTraining bool) error
}

```

owulveryck:
Hi!

Good idea indeed.
I’ve got a question. Why is the ExprGraph carrying the evalMode flag (besides commodity)?
My understanding is that the evalMode is a runtime only option; but it somehow alters the graph structure which is counter intuitive to me.

dcu:
> Hi!
> 
> Good idea indeed.
> I’ve got a question. Why is the ExprGraph carrying the evalMode flag (besides commodity)?
> My understanding is that the evalMode is a runtime only option; but it somehow alters the graph structure which is counter intuitive to me.

I was thinking about dropout since it is not a full operation so the decision can't be delayed until runtime but you're totally right eval mode is a runtime thing so it won't work anyway. I guess we should make dropout an Op

dcu:
> I'm back. Still sick and sinuses still heavily infected but I think this is somewhat a good idea though I think the interface should be one method:
> 
> ```
> type EvalModeOp interface{
>     SetTraining(isTraining bool) error
> }
> ```

I made this change. please take a look again
should I go ahead and make a dropout op?

dcu:
I just implemented Dropout as an Op, it should be faster than previous implementation

PR #494, opened 2021-07-04
Fix adam

dcu:
The problem was that it was increasing the weights twice in every step.

I verified the output against this pytorch program and it matches:
```
import torch
import numpy as np

model = torch.nn.Linear(4, 1, bias=False)
with torch.no_grad():
     model.weight.copy_(torch.ones_like(model.weight))

optimizer = torch.optim.Adam([model.weight], lr=0.1)

for i, epoch in enumerate(np.arange(0.0, 1.0, 0.1)):
    optimizer.zero_grad()
    output = model(torch.autograd.Variable(torch.tensor([epoch, epoch, epoch, epoch]).float()))
    # output.retain_grad()

    loss = torch.mean(output)
    # loss.retain_grad()

    loss.backward()
    optimizer.step()

    print("#", i, model.weight.data)
```

### DIFF
--- a/solvers.go
+++ b/solvers.go
@@ -469,7 +469,7 @@ func (s *AdamSolver) Step(model []ValueGrad) (err error) {
 			w := weights.(*tensor.Dense)
 			v := cvv.(*tensor.Dense)
 
-			var l1reg, l2reg, clip, negClip, beta1, beta2, omβ1, omβ2, eps, eta, onePerBatch interface{}
+			var l1reg, l2reg, clip, negClip, beta1, beta2, omβ1, omβ2, eps, negEta, onePerBatch interface{}
 			var correctionV1, correctionV2 interface{}
 			switch m.Dtype() {
 			case tensor.Float64:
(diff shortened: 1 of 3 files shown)


solvers.go
```diff
@@ -597,11 +597,11 @@ func (s *AdamSolver) Step(model []ValueGrad) (err error) {
 				return
 			}
 
-			if _, err = tensor.Mul(mHats, eta, tensor.UseUnsafe()); err != nil {
+			if _, err = tensor.Mul(mHats, negEta, tensor.UseUnsafe()); err != nil {
 				return errors.Wrap(err, pointWiseMulFail)
 			}
 
-			if _, err = tensor.Div(mHats, vHats, tensor.WithIncr(w)); err != nil {
```
dcu, line 604: this was the problem

chewxy: wait why is this the problem?

dcu: because the weights are added in the line 611 (couple of blocks below) and the formula says they are added once not twice 

image

PR #509, opened 2021-10-24
CTC loss

dcu:
CTC as described in this paper: http://www.cs.toronto.edu/~graves/icml_2006.pdf


Mddct:
support gpu？

chewxy:
I will add GPU support, likely by calling CuDNN's CTC function.. .but I gotta merge this one first

Mddct:
> I will add GPU support, likely by calling CuDNN's CTC function.. .but I gotta merge this one first

Awesome， If there is a GPU implementation, I plan to write an example of speech recognition

PR #510, opened 2021-11-07
Fix BN gradient calculation

dcu:
This version use auto-gradient for scale and bias instead of doing manual calculation


chewxy (approved):
LGTM

PR #519, opened 2022-03-26
Add LogSoftMax operation

dcu:


chewxy:
Interesting. So the graph replacement algorithm already does replace `Log(Softmax(x)` with a singular `softmaxOp` with `log=true`. But I think this API could be useful?

PR #536, opened 2022-10-26
Fix BinaryXent got NaN and also formula comment

auxten:
1. Fix formula in comment
2. Fix log(1-prob) with something like log(1.000001-prob), if not that the loss func will got NaN in some rare case. I have test that in my code https://github.com/auxten/edgeRec/blob/main/model/din/model.go#L58

### DIFF
--- a/const.go
+++ b/const.go
@@ -84,16 +84,18 @@ const (
 var empty struct{}
 
 var (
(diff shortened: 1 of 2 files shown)


chewxy:
Hey nice fix.

PR #544, opened 2023-03-24
Fix for CUDA stuff not compiling due to trying to use the deprecated `tensor.Memory.Pointer` function

JoshPattman:
# The Problem
So when I try to run with build tag cuda, it won't compile. It looks like the interface `Memory` in package `gorgonia.tensor` is supposed to declare a function `Pointer` (or at least this package expects it to). When I added that function to the interface it broke many things.

After further investigation, it looks like the `Pointer` method was actually removed from the tensor package. So I guess this might actually be the correct way to fix the problem.

# My Fix
I have just added a function that converts a `Value` into an `unsafe.Pointer`. This may be a bad way to do this, but it seems to work correctly for me.

# My package versions
`gorgonia v0.9.17`
`tensor v0.9.24`
`cu v0.9.4`

# Tests
I ran `go test` in the gorgonia directory and it passed. I think this problem might have slipped through the automated tests if they are not testing stuff with the `cuda` build tag. Maby this should be added?

### DIFF
--- a/cuda.go
+++ b/cuda.go
@@ -138,7 +138,7 @@ func (m *ExternMetadata) GetFromValue(dev Device, v Value) (tensor.Memory, error
 	}
 	ptr := cu.DevicePtr(mem.Uintptr())
 	ctx := m.engines[dev].Context()
-	ctx.MemcpyHtoD(ptr, v.Pointer(), memsize)
+	ctx.MemcpyHtoD(ptr, valueToPointer(v), memsize)
 	return cu.DevicePtr(ptr), nil
 }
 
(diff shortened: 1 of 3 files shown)


chewxy:
Hmm.. I'm quite sure the `tensor` package hasn't removed `Pointer`... as I'm still using it on a daily basis, with CUDA. Let me check

chewxy:
nope you  are right... I will merge this once it passes the tests

JoshPattman:
I think you might need to approve the workflow as I am a first time contributor. On my end it still says waiting for approval to run checks.

JoshPattman:
@chewxy Please may you approve the workflow so it can perform tests

chewxy:
sorry... fixed

Issue #552, opened 2023-09-14
BatchedMatMul bug when applying to 1-shape matrix


wzzhu:
The operator batchedMatMul slices the batch of matrices batch by batch. But the slice of tensor will reshaped by tensor.Slice operation to remove those dimension with 1-shape. For example, for 10x1x1 tensor A to multiply 10x1x10 matrix B. The slice of A will be reshaped to scalar, which results in error to multiplying matrix B's slice which has 1x10 shape.

e.g.,

```
func ExampleBatchedMatMul() {
	g := NewGraph()
	d := NewTensor(g, Float64, 3, WithShape(10, 1, 1), WithInit(RangedFrom(1)), WithName("d"))
	e := NewTensor(g, Float64, 3, WithShape(10, 1, 10), WithInit(RangedFrom(11)), WithName("e"))
	f, err := BatchedMatMul(d, e)
	if err != nil {
		log.Fatal(err)
	}
	m := NewTapeMachine(g)
	if err := m.RunAll(); err != nil {
		log.Fatal(err)
	}
}
```

PR #555, opened 2023-09-14
 Fix BatchedMatMul bugs

wzzhu:
Fixed https://github.com/gorgonia/gorgonia/issues/552
Three things in BatchedMatMul function included.

When BatchedMatMul is performed on matrix with (1, 1) shape, it will return error as the Slice operation will return a scalar as the operand. This will cause matrix multiplication requirements invalid.

Originally if the matrix C for result of matrix A x B is nil, its shape will be set to pre-defined A.row and B.col. But if transpose flag is set for either A or B, the matrix for C will be set incorrectly as it doesn't consider the transposition.

Fix incorrect newShape for matrix C re-uses the slice backing data by adding Clone for newShape.

### DIFF
--- a/example_linalg_test.go
+++ b/example_linalg_test.go
@@ -23,6 +23,17 @@ func ExampleBatchedMatMul() {
 	if err != nil {
 		log.Fatal(err)
 	}
+	h, err := BatchedMatMul(e, d, true, true)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


chewxy:
LGTM. Merged. I've also sent you an email @wzzhu

